Issue #4, opened 2017-05-04
Busy waiting


Lercher:
in src\JsonRpc\OutputHandler.cs there is a tight loop plus a Delay/Sleep() which looks quite inefficient to me. I've prepared a pull request for this, you might want to look into.

There is a similar one in src\JsonRpc\InputHandler.cs at `private async void ProcessRequestQueue()`, but I'm afraid I'm to dumb to understand, what is done there. So could you please evaluate, if you could probably change the  `ConcurrentQueue<>` to a `BlockingCollection<>` plus a `CancellationTokenSource`here as well?

Thanks
Martin

david-driscoll:
The call to `Delay` / `Sleep` only happens when the Queue has been drained.  If there are still pending requests / responses in the queue then there will no wait.  The intent is to avoid having a few threads constantly running in circles waiting for events to be received.  If there is a better implementation, I'm totally open to using alternatives.

------

The intent in `ProcessRequestQueue` is to avoid processing requests strictly in serial.  `IRequestProcessIdentifier` allows the implementer to decide what can be done in parallel, and what must be done in serial.

For example a text change event will almost always be blocking, as it will affect any of the next requests that happen.  However something like autocomplete, or signature help are not bound in that way.  This way those requests can be served back to the editor as fast as possible.

Strictly speaking the protocol says all requests must respond in order received, but this allows an implementer the ability to bypass some of that for endpoints that don't affect the overall solution.

Note: I haven't started a real integration some of these implementations will likely change.

Lercher:
(1) on "The call to ...": I only agree partly. The queue is almost always empty or seldom contains one item, so assuming a configured delay of 50ms, the old consumer thread burns some cycles, only to check the queue is empty, waits 50ms, repeat and drains the battery. Now assume a unit of work arrives in the middle of the 50ms delay, then for 25ms plus thread switching time nothing happens, except delay. The new code should not burn CPU cycles, if there is nothing to do. And it has no lag besides task switching time. If the queue is longer than one item, all work is done at top speed, just as before. 

Thanks for accepting my fix.

Lercher:
(2) on "The intent in ...": Before I feel confident enough to offer you a change, could you please verify/comment on these statements:
* In general: all units of work are scheduled in parallel and started in parallel, except ...
* ... an item in `_queue` marked as `Serial` waits for all previously scheduled units of work to be finished, plus ...
[...]

Issue #3, opened 2017-05-03
UTF-8


Lercher:
My Console.In and .Out Encodings in the running SampleServer are utf-8 and the spec says ASCII in the header part and usually utf-8 in the body. So I think, converting a char[] with Encoding.ASCII.GetString(Encoding.ASCII.GetBytes(buffer)) is probably a bad idea. Please consider changing it in `src\JsonRpc\InputHandler.cs`:

````
    var headersContent = new string(buffer, 0, current);
    // var headersContent = Encoding.ASCII.GetString(Encoding.ASCII.GetBytes(buffer, 0, current));
````

which works because each ASCII byte is the same in UTF-8, and

````
    var payload = new string(requestBuffer);
    // var payload = Encoding.ASCII.GetString(Encoding.ASCII.GetBytes(requestBuffer));
````

so that the TextReader can do the hard work of decoding the encoding.

Martin

Lercher:
should be resolved now.

PR #5, opened 2017-05-04
Fix for Issue #3 and #4

Lercher:
Sorry, I'm not familiar with pull requests and so I wasn't able to provide separate ones per issue.

Additionally you might want to call Dispose() in test/JsonRpc.Tests/OutputHandlerTests.cs, but that looked to invasive to me.

Best regards, Martin

### DIFF
--- a/src/JsonRpc/InputHandler.cs
+++ b/src/JsonRpc/InputHandler.cs
@@ -87,7 +87,7 @@ namespace JsonRpc
                     current += await _input.ReadBlockAsync(buffer, current, 1);
                 }
 
-                var headersContent = Encoding.ASCII.GetString(Encoding.ASCII.GetBytes(buffer, 0, current));
+                var headersContent = new string(buffer, 0, current);
                 var headers = headersContent.Split(HeaderKeys, StringSplitOptions.RemoveEmptyEntries);
                 long length = 0;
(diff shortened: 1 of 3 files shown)


src/JsonRpc/OutputHandler.cs
```diff
@@ -9,42 +9,38 @@ namespace JsonRpc
...
+
         private void ProcessOutputQueue()
         {
+            var token = _cancel.Token;
             while (true)
             {
                 if (_thread == null) return;
-
-                if (_queue.TryDequeue(out var value))
+                if (_queue.TryTake(out var value, -1, token))
```
david-driscoll: To make it easier to understand what `-1` means, we should probably change it to `Timeout.Infinite`, which is a little more transparent about what it's trying to do.

src/JsonRpc/InputHandler.cs
```diff
@@ -104,7 +104,7 @@ public void Start()
 
                 await _input.ReadBlockAsync(requestBuffer, 0, requestBuffer.Length);
 
-                var payload = Encoding.ASCII.GetString(Encoding.ASCII.GetBytes(requestBuffer));
+                var payload = new string(requestBuffer);
```
david-driscoll: The more you know... 😄

PR #7, opened 2017-05-06
Fix failing test

Lercher:
Sorry for the trouble. I noticed not only until today, that my VS2017 only ran the Lsp.Tests but not the JsonRpc ones. I overlooked (1) that cancelling the token causes an ex when canceled and (2) that the CancellationTokenSource needs to be Disposed as well.

Martin

### DIFF
--- a/src/JsonRpc/OutputHandler.cs
+++ b/src/JsonRpc/OutputHandler.cs
@@ -19,7 +19,8 @@ namespace JsonRpc
             _output = output;
             _queue = new BlockingCollection<object>();
             _cancel = new CancellationTokenSource();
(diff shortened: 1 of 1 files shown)


Lercher:
Fails in `Coverage` at `C:\projects\csharp-language-server-protocol\test\JsonRpc.Tests\InputHandlerTests.cs(202,0): at JsonRpc.Tests.InputHandlerTests.ShouldHandleResponse()`

Didn't change that, what did I wrong?

david-driscoll:
The new `throw` scenario isn't hit, hence the coverage.  It's nothing drastic however. 👍

PR #12, opened 2017-05-09
BugFix: content-length is counted in bytes

Lercher:
Hence we can't use TextReaders and TextWriters in In/OutputHandlers, because otherwise reading chars when there is only a bytes count would be really hard. So I converted everything to streams plus UTF8 Encoding methods plus some test enhancements.

This change is _important_, because otherwise there is _no_ parsing of text containing non-ASCII characters, which is quite common outside of the US ;-) With it, I can now parse a real Coco/R grammar with a real example :-)

Note: This fix is based on the _current_ OmniSharp/master. As it is some fixes behind, I expect some manual merging.

Martin

### DIFF
--- a/sample/SampleServer/Program.cs
+++ b/sample/SampleServer/Program.cs
@@ -24,7 +24,7 @@ namespace SampleServer
             //    await Task.Delay(100);
             //}
 
-            var server = new LanguageServer(Console.In, Console.Out);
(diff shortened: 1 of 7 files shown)


david-driscoll:
@Lercher merged all the other changes, there are a few minor differences to make here, then we're good to go.  👍

Lercher:
Did the expected manual merges

david-driscoll:
👍

PR #14, opened 2017-05-11
ProcessInputThread should be sync

Lercher:
Since you allocate an exclusive background thread for reading, it should be used in that way. Fix+Test.

Martin


david-driscoll:
👍

PR #20, opened 2017-09-23
Skip dynamic registration of server capabilities if client does not support it

tintoy:
As noted by LSP protocol docs, clients do not have to support this:

https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#new-register-capability

And if the client does not support this, it currently causes LanguageServer.Initialize to hang (since the resulting exception is not handled correctly).


david-driscoll:
lgtm!

Issue #22, opened 2017-10-05
Language server's initial response to "initialize" request does not correctly report server capabilities


tintoy:
Hi.

As noted [here](https://github.com/tintoy/msbuild-project-tools-vscode/issues/12#issuecomment-334449018) the language server is not correctly reporting its capabilities when responding to the `initialize` request (although _if_ the client supports dynamic registration then the subsequent capability-registration message will correctly report its capabilities).

I believe I've tracked this down to the use of `LanguageServer.HasHandler<T>`, which is never going to behave as expected because the interface captured in `HandlerCollection` is going to be `IRequestHandler<HoverParams, Hover>` not `IHoverProvider`:

https://github.com/OmniSharp/csharp-language-server-protocol/blob/4bc2a15e43593e6e459e9b0784cb85fadcf89c34/src/Lsp/LanguageServer.cs#L130

I believe this could be fixed by either capturing the original interface (`IHoverHandler`) when adding the handler to the collection, finding the underlying `IRequestHandler<HoverParams, Hover>` interface and matching on that instead, or simply matching on JSON-RPC method name (personally, this seems like the most robust and easy-to-reason-about option).

For example:

```csharp
bool HasHandler<T>()
    where T : IJsonRpcHandler
{
    var method = LspHelper.GetMethodName<T>();

    return _collection.Get(method).Any();
}
```

I'm happy to open a PR if you don't have time to fix this, but if so let me know which solution you'd prefer.

david-driscoll:
Sorry didn't see this the other day.   I've changed it to `_collection.Any(z => z.Handler is T);` with the next release (shortly) which hopefully will solve the problem.  If that doesn't then we can move to use the method names.

tintoy:
No worries - thanks for having a look at it!

BTW, might it be worth considering a match on JSON-RPC method name? Because (although this is admittedly an edge case) there's nothing to indicate to an API consumer that creating their own handler interface with the correct `[Method]` attribute _will_ correctly handle the messages but _won't_ result in a correct capability declaration.

david-driscoll:
That's a fair point.  My main goal was to just have things be automatic, and I (generally) dislike strings in C# because they're easy to mess up or feel kludgy to get right.

We'd have to create a const string field for each item we wanted to add, and then use that anywhere the text is used.  That's why I was using the interface to basically remove needing the string, it's only marked one place, implement the interface and you're good to go.

[...]

PR #26, opened 2017-10-10
Reacting to latest changes for LSP

david-driscoll:
Updated to the latest changes of https://github.com/Microsoft/language-server-protocol.

Also added a few extension methods for logging and such.


david-driscoll:
@tintoy want to sanity check? 🎉

tintoy:
Hey, nice! Been looking forward to seeing commit characters :)

tintoy:
It looks like the same logging methods are declared in 2 different classes?

david-driscoll:
Theres log methods and show methods, since they're both similar I figured might as well extend both... but yes I didn't rename them all... yay...

tintoy:
Good to have them though :)

tintoy:
BTW, will create a separate issue, but it looks like the server doesn't entirely handle clients that only support static registration of server capabilities. Will put together a repro, but it looks easy to fix :)

tintoy:
> it looks like the server doesn't entirely handle clients that only support static registration of server capabilities

Oops, just realised - I forgot I'd already reported this under #22 :)

david-driscoll:
@tintoy made a small change for that, once builds pass I'll merge and make a new release if you want to try it out.

tintoy:
> made a small change for that, once builds pass I'll merge and make a new release if you want to try it out.

Thanks!

Was just wondering - might it be a little more semantically clear to rely on _matching_ on JSON-RPC method name rather than handler type? The only reason I ask is that (although this is admittedly an edge case) there's nothing to indicate to an API consumer that creating their own handler interface with the correct `[Method]` attribute _will_ correctly handle the messages but _won't_ result in a correct capability declaration.

PR #36, opened 2017-10-17
LspRouter update, update `Supports<>`

### LINKED ISSUE #35: Conversion of unsupported `ClientCapabilities` is incompatible with other LSP implementations.
Client capabilities in the LSP are usually specified as an optional wrapper around a boolean value, e.g.  the `workspaceEdit` capability in `WorkspaceClientCapabilities` is a wrapper around a `documentChanges` boolean. The possible values here are: wrapper around `true`, wrapper around `false`, or a non-existent or `null` value to indicate the capability is not supported/recognised.

I observe that the [default behaviour of the `SupportsConverter`](https://github.com/OmniSharp/csharp-language-server-protocol/blob/master/src/Lsp/Converters/SupportsConverter.cs#L38) is to always provide the value `false` instead of `null` when the capability is not supported. This makes the OmniSharp protocol implementation incompatible with language server implementations that expect an object wrapper around a boolean (instead of just a boolean) in capability fields such as `WorkspaceClientCapabilities.workspaceEdit`.

The example where I observed this incompatibility used [Eclipse LSP4J](https://github.com/eclipse/lsp4j) for the server, and the [.NET language client](https://github.com/tintoy/dotnet-language-client) (which relies on OmniSharp for the protocol implementation) as the client.

david-driscoll:
Two things:
* Updated the router to pick the proper `ITextDocumentSyncHandler` previously this would pick the first one, which was only right if the one you wanted was first.  No other handlers would.
* `ITextDocumentSyncHandler.GetTextDocumentAttributes` can return null, previously this was assumed to be not null.
  * However one of your handlers should return a non null value or nothing will get routed.
* Added some tests around `LspRequestRouter` to ensure it's working as intended (I hope, 😱)
* Fixes #35 using the hacky feeling `ShouldSerializeX` in `Newtonsoft.Json`.  There is be a better way I'm sure, but these methods will be hidden from intellisense for consumers.


david-driscoll:
@tintoy going to push this as `0.4.0`, let me know if you run into any problems.

tintoy:
Cheers! Will give it a go in an hour or so

tintoy:
What's the expected behaviour for requests / notifications where multiple handlers for the same method are registered?

tintoy:
Ah, needs a logger factory now; will add one.

tintoy:
Yep! Works fine, cheers :)

tintoy:
Oops, spoke too soon:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at OmniSharp.Extensions.LanguageServer.LanguageServer.HasHandler[T](DynamicCapability capability)
   at OmniSharp.Extensions.LanguageServer.LanguageServer.<OmniSharp-Extensions-JsonRpc-IRequestHandler<OmniSharp-Extensions-LanguageServer-Models-InitializeParams\,OmniSharp-Extensions-LanguageServer-Models-InitializeResult>-Handle>d__26.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at OmniSharp.Extensions.LanguageServer.LspRequestRouter.<RouteRequest>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at OmniSharp.Extensions.JsonRpc.InputHandler.<>c__DisplayClass17_0.<<HandleRequest>b__3>d.MoveNext()
```
As shown here:

https://github.com/tintoy/dotnet-language-client/blob/master/samples/SingleProcess/Program.cs#L173

tintoy:
Looks like the overload that takes `DynamicCapability` instead of `SupportsDynamicCapability<T>` needs to do a null-check; IIRC, it's legal for those fields to be `null`.

david-driscoll:
I didn't get to this last night... fixing shortly.

PR #37, opened 2017-10-18
Updated capabilities to indicate which handler they are tied too, th…

david-driscoll:
…is helps update the HasHandler logic.  In addition extracted HasHandler out into a unit testable class.


david-driscoll:
Someday I'll fix travis...

PR #38, opened 2017-10-19
Breaking: Error object was misshaped, textdocumentsync double registered.

david-driscoll:
* `Error` didn't have the correct property attached to it.  Renamed class to `RpcError` and updated property name to be `Error`
* Fixed `ErrorResponse` to use two derived types, this is to ensure that both the `result` and `error` properties will be set correctly.
* Fixed an problem where `ITextDocumentSyncHandler` was being registered twice for servers that support dynamic registration.

deploying as `v0.5.0` cc @tintoy 

I tested with a ProjectReference so things appear to be working in my project anyway.


tintoy:
Cheers - `0.5.0` is not indexed on NuGet yet so I can't install it, but I _can_ confirm that `0.4.1` fixed the original problem (`ServerCapabilities.HoverProvider` is  now correctly returned). Will give `0.5.0` a try when it becomes available.

david-driscoll:
Looks to be indexed now.

tintoy:
After [upgrading](https://github.com/tintoy/dotnet-language-client/commit/4084704d990eda5567a7a750160a4ffab5054ab2) to `v0.5.0`, my [single-process example](https://github.com/tintoy/dotnet-language-client/blob/4084704d990eda5567a7a750160a4ffab5054ab2/samples/SingleProcess/Program.cs) now crashes when shutting down with an unhandled TaskCanceledException. Works correctly when I roll back to `v0.4.1`, BTW. Nothing obvious looks amiss, and given the stack trace for the TaskCancelledException, it looks like it might be an unobserved task result?

```
System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.IO.Pipes.PipeStream.Write(Byte[] buffer, Int32 offset, Int32 count)
   at OmniSharp.Extensions.JsonRpc.OutputHandler.ProcessOutputQueue()
   at System.Threading.Thread.ThreadMain_ThreadStart()
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
```

tintoy:
Hooray for verbose logging!

```
[06:32:42 DBG] Unable to find shutdown, methods found include initialize:OmniSharp.Extensions.LanguageServer.LanguageServer, initialized:OmniSharp.Extensions.LanguageServer.LanguageServer, $/cancelRequest:OmniSharp.Extensions.LanguageServer.Handlers.CancelRequestHandler, workspace/didChangeConfiguration:Common.ConfigurationHandler, textDocument/hover:Common.HoverHandler, dummy:Common.DummyHandler
```

david-driscoll:
a crash on shutdown is the only way to shutdown I'm tell you!

david-driscoll:
At least the logging is useful... time to add some extra tests.

tintoy:
Yeah for sure - logging saved my bacon several times while I was building the client :-D

PR #43, opened 2017-11-03
Initial port of LSP client from tintoy/dotnet-language-client.

tintoy:
Hey - as discussed in OmniSharp/csharp-language-server-protocol#42, I've pulled everything into the `OmniSharp.LanguageServerProtocol.Client` namespace. Have tried to change as little as possible as part of the port so we can talk it over in its present state.

BTW, the client library has to target `netstandard2.0` (`netstandard1.6` doesn't have the required support for Pipes).


tintoy:
It looks like CI may be failing due to an unrelated issue?

```
Error finding repository for 'https://ci.appveyor.com/nuget/csharp-language-server-protocol': An error occurred while retrieving package metadata for 'xunit.runner.console.2.3.0' from source 'AppVeyorProjectFeed'.
```

tintoy:
Ah, no, I see now. Need the `dotnet-xunit` CLI tool.

david-driscoll:
looking at this today, sorry I completely forgot about it this weekend.

src/Client/Client.csproj
```diff
@@ -0,0 +1,21 @@
...
+  <PropertyGroup>
+        <TargetFrameworks>netstandard2.0</TargetFrameworks>
+        <PlatformTarget>AnyCPU</PlatformTarget>
+        <AssemblyName>OmniSharp.Extensions.LanguageClient</AssemblyName>
+        <RootNamespace>OmniSharp.Extensions.LanguageServerProtocol.Client</RootNamespace>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
+        <PackageReference Include="Serilog" Version="2.5.0" />
```
david-driscoll: I'm hesitant to depend on `Serilog` by itself here, as I would generally prefer to not prescribe logging frameworks for consumers.

However Serilog is a great logger, so I'm okay with it for now, and perhaps refactor it away (for MS.Extensions) in a future (breaking) change.

src/Client/Clients/TextDocumentClient.Completions.cs
```diff
@@ -0,0 +1,64 @@
+﻿using OmniSharp.Extensions.LanguageServer.Models;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace OmniSharp.Extensions.LanguageServerProtocol.Client.Clients
+{
+    using Utilities;
```
david-driscoll: _nit_: Not a huge fan of nested usings.

src/Client/Clients/TextDocumentClient.Completions.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// </param>
+        /// <returns>
+        ///     A <see cref="Task{TResult}"/> that resolves to the completions or <c>null</c> if no completions are available at the specified position.
+        /// </returns>
+        public Task<CompletionList> Completions(string filePath, int line, int column, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (String.IsNullOrWhiteSpace(filePath))
+                throw new ArgumentException($"Argument cannot be null, empty, or entirely composed of whitespace: {nameof(filePath)}.", nameof(filePath));
+
+            return PositionalRequest<CompletionList>("textDocument/completion", filePath, line, column, cancellationToken);
```
david-driscoll: _future_: Now that the strings are showing up in more places we should probably add a static constants class somewhere.  I'll add an issue for this.

src/Client/Clients/TextDocumentClient.Completions.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <param name="column">
+        ///     The target column (0-based).
+        /// </param>
+        /// <param name="cancellationToken">
+        ///     An optional <see cref="CancellationToken"/> that can be used to cancel the request.
+        /// </param>
+        /// <returns>
+        ///     A <see cref="Task{TResult}"/> that resolves to the completions or <c>null</c> if no completions are available at the specified position.
+        /// </returns>
+        public Task<CompletionList> Completions(Uri documentUri, int line, int column, CancellationToken cancellationToken = default(CancellationToken))
```
david-driscoll: We could probably make this a wrapper for the other method, and just pass documentUri as a string or better yet reverse it so that we pass `filePath` into this method.  That would ensure that the string is a proper `Uri`.

src/Client/HandlerDelegates.cs
```diff
@@ -0,0 +1,89 @@
...
+
+namespace OmniSharp.Extensions.LanguageServerProtocol.Client
+{
+    /// <summary>
+    ///     A handler for empty notifications.
+    /// </summary>
+    /// <returns>
+    ///     A <see cref="Task"/> representing the operation.
+    /// </returns>
+    public delegate void EmptyNotificationHandler();
```
david-driscoll: Should be safe to rename to `NotificationHandler`.

src/Client/HandlerDelegates.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// <summary>
+    ///     A handler for notifications.
+    /// </summary>
+    /// <typeparam name="TNotification">
+    ///     The notification message type.
+    /// </typeparam>
+    /// <param name="notification">
+    ///     The notification message.
+    /// </param>
+    public delegate void NotificationHandler<TNotification>(TNotification notification);
```
david-driscoll: I recently changed these on the server to be `Task` instead of `void` to allow them to block processing until the operation completes, this ensures things like text updates are completed.

Question: Should we do the same here for the client?  (I honestly don't know)

src/Client/Handlers/IHandler.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        /// <summary>
+        ///     The name of the method handled by the handler.
+        /// </summary>
+        string Method { get; }
+
+        /// <summary>
+        ///     The kind of handler.
+        /// </summary>
+        HandlerKind Kind { get; }
```
david-driscoll: `HandlerKind` and the `Kind` property seem unnecessary.  Worst case we can pattern match, but currently they aren't used anywhere but setting them..

src/Client/Protocol/LspConnection.cs
```diff
@@ -0,0 +1,1011 @@
...
+            {
+                Log.Warning("Unable to dispatch incoming {RequestMethod} request {RequestId} (no handler registered).", requestMessage.Method, requestId);
+
+                _outgoing.TryAdd(
+                    new JsonRpcMessages.MethodNotFound(requestMessage.Id, requestMessage.Method)
+                );
+            }
+
+#pragma warning disable CS4014 // Continuation does the work we need; no need to await it as this would tie up the dispatch loop.
+            handlerTask.ContinueWith(_ =>
```
david-driscoll: This will throw if `handlerTask` is null, should probably `return` in the above if.

tintoy:
Hi - I'll sort all these out over the next couple of days, thanks for taking the time to do a detailed review :)

src/Client/Client.csproj
tintoy: Thanks for the reminder - I'm fine with changing over to MEL for logging, I just forgot about it :)

I'll modify them to take an `ILogger` / `ILoggerFactory` instead (some components need to pass a different `ILogger<TComponent>` to child components).

src/Client/Clients/TextDocumentClient.Completions.cs
tintoy: No problem; I tend to use it as a reminder of which namespaces are from "my" code vs external code, but I'm happy to change this over. My rule is usually "match the style already used by the project" :)

src/Client/Clients/TextDocumentClient.Completions.cs
tintoy: Yeah, I probably should have done this in the first place (even when my project was stand-alone) but haven't got around to it.

src/Client/Clients/TextDocumentClient.Completions.cs
tintoy: Yep, good catch :)

src/Client/HandlerDelegates.cs
tintoy: Yeah, I was trying to make it explicit to consumers that empty notifications are not the same as notifications with payloads (the language server implementation will freak out if you don't pass a payload when one is expected.

But I agree it makes more sense to just vary by type parameters.

src/Client/HandlerDelegates.cs
tintoy: Interesting point; I think that makes sense server-side, but I reckon _sending_ a notification should be fire-and-forget. I can't think of a reason that a client would want to block until it's been sent.

src/Client/Handlers/IHandler.cs
tintoy: My original thinking on this was that the handlers themselves would be generic and so it would be hard to pattern-match on them without already knowing the handler payload type (but you're right it's probably redundant now).

src/Client/Protocol/LspConnection.cs
tintoy: Hah! Yes, good catch :)

src/Client/HandlerDelegates.cs
tintoy: I've changed this, but I'd like to keep the underlying `TryHandleEmptyNotification` methods where there are no generic type parameters because I got caught more than once by accidentally calling the wrong overload.

tintoy:
Ok - the only thing left to do (I think) is the switch from Serilog to MEL. That may take a little longer; Just need to make sure I'm not directly using Serilog-specific functionality anywhere and I'll need to rework the test framework to handle this (but that should be doable).

tintoy:
I'll also see what I can do to bump up the test coverage.

tintoy:
Logging switched over, too - that turned out to be easier than I expected :grin:

david-driscoll:
FYI For logging I'm totally fine you bring in Serilog and MEL to handle logging into Xunit, it just makes things so much easier.

I would just prefer that the external interface tie to MEL if possible.

src/Client/HandlerDelegates.cs
david-driscoll: and thats why I ask the question. :D

tintoy:
No worries - turns out it was easy to use MEL for in-test logging (see the `TestBase` class for details).

tintoy:
We could probably also create an `ILoggerProvider` to log to the language server via `window/logMessage` :)

tintoy:
But if you want the Serilog logging stuff for Xunit, I'm happy to contribute that too (although it's no longer needed here).

david-driscoll:
If it works we'll leave it alone for now. 👍

tintoy:
:wink:

test

david-driscoll:
Hey Adam, were you going to try and bring up the coverage anymore?  If not I'm fine merging this as is, and we can add additional coverage later.

tintoy:
Actually given all the other stuff going on this morning, perhaps you can merge it and I'll open a new PR for the additional tests sometime tonight (at conference today)?

tintoy:
Woohoo!

PR #50, opened 2017-11-30
[Breaking] Split protocol into it's own project

david-driscoll:
* This makes a new package that is used for both client and server
  * this also decouples the client from the server.
* Changed out all the method names `textDocument/x` etc to all be static strings.
  * I feel like there are some more ways we can improve this experience for clients and servers by creating a set of base classes and some generic type magic, similar to how the vscode lsp implementation does... however I do not want to tackle that here.


david-driscoll:
cc @RLittlesII

RLittlesII:
I would generally agree with your reflective magic as static strings are an eye sore, and further agree it is a separate body of work.  The refactoring on this looks good.

PR #51, opened 2017-12-01
[WIP] bring up to date with latest spec

david-driscoll:
Things left to do...
- [x] Fix any build errors
- [x] Unit Tests of the changed bits
- [x] Unit Test the new bits
- [x] Synchronize the `IXOptions` interfaces with capabilities

Things to figure out
- [ ] How do we want to enforce errors for things that are not supported by the old protocol?
  * In this case `MarkedString` was deprecated but it should still be supported for LSP 2.0 and existing LSP clients.  We have `contentFormat` that we can run off of.
  * It feels like we'll have to refactor things and deal with instances of `JsonSerializer` or `JsonSerializerSettings` manually.  Today we cheat and just use the static methods.  With this latest change we have to adapt our serialization to what the editor supports.


david-driscoll:
Fair point I'll add a wrapper interface.

david-driscoll:
cc @tintoy if you get a chance to look it over... 

For serialization in the client I had to use a singleton instance of the new Serialzier, this needs to be fixed.  Other than that all the serialization logic is now coupled to the Contract Resolver and Converter classes.

tintoy:
I notice the empty notification handler delegate was removed - are notifications without payloads no longer supported?

src/Client/Handlers/DelegateNotificationHandler.cs
```diff
@@ -50,7 +52,7 @@ public DelegateNotificationHandler(string method, NotificationHandler<TNotificat
             await Task.Yield();
 
             Handler(
-                notification != null ? notification.ToObject<TNotification>() : default(TNotification)
+                notification != null ? notification.ToObject<TNotification>(Serializer.Instance.JsonSerializer /* Fix me: this is ugly */) : default(TNotification)
```
tintoy: Hmm, I'm thinking maybe we could make the the `JsonSerializerSettings` assignable via a property on `ICustomizeJsonSerializer` (new interface with a `SerializerSettings` property that `INotificationHandler` and friends could inherit from)? Either that or pass it into the `Invoke()` methods but the property seems cleaner.

david-driscoll: What if we change `Invoke` to take an `object` instead and deserialize the value before we call `Invoke`?

tintoy: Yeah that would be simpler - nice one :)

tintoy: Mind you, how do you know which type to deserialise into? I think that's why I did it in the handler because the parameter type is known.

tintoy: I suppose the handler could expose a `ParametersType` / `PayloadType` property?

david-driscoll:
I think I'm going to land this for now, and we can come back to the serialization problem.  The behavior hasn't changed, the code just isn't as nice.

Let me know what you think.

tintoy:
Sure works for me :)

tintoy:
Want me to open a PR for fixing up the client serialiser stuff?

david-driscoll:
Getting back to things after the holidays, everything is passing, going to merge, let me know if you see any problems!

Issue #54, opened 2017-12-23
Odd log output from TextDocumentMatcher


tintoy:
Here's an example log entry:

![logentry](https://user-images.githubusercontent.com/1646284/34322640-345e4eda-e881-11e7-903e-8aa14d853eee.jpg)

Note how the handler interface is simply `System.RuntimeType` - this seems a little odd, and looking at the code I think it's because we're calling `GetType().FullName` when all we need is `.FullName`.

PR #55, opened 2017-12-23
Log correct type for handler interface.

tintoy:
Removed call to `GetType()` on something that's already a `Type` :)

Fixes OmniSharp/csharp-language-server-protocol#54.


tintoy:
BTW, this logging has already helped me track down a long-standing bug in msbuild-project-tools-vs-code :)

PR #57, opened 2018-01-10
Added check to allow non spec compliant consumers to send us null for…

david-driscoll:
… params.  When a null is given we won't error, instead we will just pretend we never got any params at all.

cc @tintoy

### DIFF
--- a/src/JsonRpc/Reciever.cs
+++ b/src/JsonRpc/Reciever.cs
@@ -79,9 +79,16 @@ namespace OmniSharp.Extensions.JsonRpc
             }
 
             var hasParams = request.TryGetValue("params", out var @params);
-            if (hasParams && @params?.Type != JTokenType.Array && @params?.Type != JTokenType.Object)
+            if (hasParams && @params?.Type != JTokenType.Array && @params?.Type != JTokenType.Object && @params?.Type != JTokenType.Null)
             {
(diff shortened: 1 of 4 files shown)


src/JsonRpc/Reciever.cs
```diff
@@ -79,11 +79,18 @@ protected virtual Renor GetRenor(JToken @object)
...
             var hasParams = request.TryGetValue("params", out var @params);
-            if (hasParams && @params?.Type != JTokenType.Array && @params?.Type != JTokenType.Object)
+            if (hasParams && @params?.Type != JTokenType.Array && @params?.Type != JTokenType.Object && @params?.Type != JTokenType.Null)
             {
                 return new InvalidRequest(requestId, "Invalid params");
             }
 
+            // Special case params such that if we get a null value (from a non spec compliant system)
+            // that we don't fall over and throw an error.
+            if (@params?.Type == JTokenType.Null)
```
tintoy: Nice catch - I would have forgotten to do this bit :)

david-driscoll:
gotta make sure linux is happy then I'll push out as a new version.

src/JsonRpc/Reciever.cs
david-driscoll: Yeah we don't want to try to deserialize the value.

PR #60, opened 2018-01-10
Enable handling of requests with no parameters.

tintoy:
An example of this is the "shutdown" request. It has no parameters (parameter type is void), and so this was failing with a swallowed exception.

Also, log the exception so if there turn out to be any other edge cases they'll be easier to diagnose.

### DIFF
--- a/src/Server/LspRequestRouter.cs
+++ b/src/Server/LspRequestRouter.cs
@@ -121,10 +121,12 @@ namespace OmniSharp.Extensions.LanguageServer.Server
                     object @params;
                     try
                     {
-                        @params = request.Params.ToObject(descriptor.Params, _serializer.JsonSerializer);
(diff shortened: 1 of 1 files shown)


src/Server/LspRequestRouter.cs
```diff
@@ -121,10 +121,12 @@ private ILspHandlerDescriptor FindDescriptor(string method, JToken @params)
                     object @params;
                     try
                     {
-                        @params = request.Params.ToObject(descriptor.Params, _serializer.JsonSerializer);
+                        @params = request.Params?.ToObject(descriptor.Params, _serializer.JsonSerializer);
```
tintoy: BTW, am wondering if perhaps `descriptor.Params` should be `descriptor.ParamsType`?

tintoy:
Ok will add a test for this :)

PR #65, opened 2018-01-31
Fixed an issue where handler descriptors would fail to be created of …

david-driscoll:
…a given handler implemented two or more intefaces. Where one interface has IRegistartion<> and the other does not.

cc @mholo65

### DIFF
--- a/src/Server/HandlerCollection.cs
+++ b/src/Server/HandlerCollection.cs
@@ -71,12 +71,12 @@ namespace OmniSharp.Extensions.LanguageServer.Server
             }
 
             var key = "default";
-            if (handler is IRegistration<TextDocumentRegistrationOptions>)
+            // This protects against the case where class implements many, possibly conflicting, interfaces.
+            if (registration != null &&
+                typeof(TextDocumentRegistrationOptions).GetTypeInfo().IsAssignableFrom(registration) &&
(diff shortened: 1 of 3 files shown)


src/Server/HandlerCollection.cs
```diff
@@ -71,12 +71,12 @@ private HandlerDescriptor GetDescriptor(string method, Type implementedType, IJs
...
+            // This protects against the case where class implements many, possibly conflicting, interfaces.
+            if (registration != null &&
+                typeof(TextDocumentRegistrationOptions).GetTypeInfo().IsAssignableFrom(registration) &&
+                handler is IRegistration<TextDocumentRegistrationOptions> handlerRegistration)
             {
-                if (GetTextDocumentRegistrationOptionsMethod
-                    .MakeGenericMethod(registration)
-                    .Invoke(handler, new object[] { handler }) is TextDocumentRegistrationOptions options)
-                    key = options.DocumentSelector;
+                key = handlerRegistration.GetRegistrationOptions()?.DocumentSelector ?? key;
```
david-driscoll: this might cause some interesting issues with some of the other derived types... I'll have to add a few more test cases to make sure won't fall over.

david-driscoll: I added some additional tests to make sure... so far so good I think.

PR #68, opened 2018-02-02
Updated to fix an issue where routing wouldn't work correctly

david-driscoll:


david-driscoll:
I have a feeling we will still have a problem with the resolve calls, but this should fix the calls to code lens themselves.

mholo65:
@david-driscoll yup. this resolved call to CodeLensHandler, but the resolve calls don't work.

PR #70, opened 2018-02-05
Route handlers for `*/resolve` endpoints

david-driscoll:
* Added `ICanBeResolved` interface
* Made a few primitives equatable
  * Range, Location, Position
* Changed `IHandlerMatcherCollection` to allow any object type
* Added `IHandlerPostProcessor` and `IHandlerPostProcessorMatcher`
* Added `ResolveCommandMatcher`
  * This matcher deals with the new `ICanBeResolved` interface
  * Calls to the actual endpoint will have `Data` wrapped in a `JObject` that looks like `{ handleType: '<type.FullName>'; data: <existing data> }`
  * Calls to `/resolve` will pull the handler type and restore the data.  No need to change the property type to `JObject` data can be a primitive, array or object.
* Added `DocumentLinkCommandMatcher`
  * This matcher deals with `documentLink/resolve` since it does not have a data field 😢 
  * This will keep a reference to the last descriptor that ran for `textDocument/documentLink`, and any new calls to `documentLink/resolve` will resolve to that handler.

cc @tintoy


david-driscoll:
I need to add some more unit tests to this, but otherwise I like the direction that's going.

david-driscoll:
I also don't like how the data field is updated during `FindHandler` I think I need to add a new interface `IHandlerPreProcessor`.

(I swear I'll rewrite this to use MediatR in the future...)

tintoy:
Cool! Will give this a once-over shortly :)

tintoy:
Just wondering - might it be simpler to just call each handler with `CanResolve(completionItem)` and let the implementer decide whether or not it's one of theirs? Just wondering if we're skirting the edge of the spec a bit here...

david-driscoll:
You raise a good point it's entirely possible that the resolve handler may not be the same as the initial handler...

I'll have to think about that.

src/Protocol/Models/CompletionList.cs
```diff
@@ -9,14 +9,22 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol.Models
     /// Represents a collection of [completion items](#CompletionItem) to be presented
     /// in the editor.
     /// </summary>
-    public class CompletionList
+    public class CompletionList : Container<CompletionItem>
```
david-driscoll: This is a breaking change unfortunately, but it isn't terrible... completion list should be supported as a simple list.

src/Server/Matchers/DocumentLinkCommandMatcher.cs
```diff
@@ -0,0 +1,53 @@
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+using OmniSharp.Extensions.LanguageServer.Protocol;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+using OmniSharp.Extensions.LanguageServer.Server.Abstractions;
+
+namespace OmniSharp.Extensions.LanguageServer.Server.Matchers
+{
+    public class DocumentLinkCommandMatcher : IHandlerMatcher, IHandlerPostProcessorMatcher, IHandlerPostProcessor
```
david-driscoll: This guy needs tests, I just haven't added them yet.

david-driscoll:
Pretty happy with the changes for the most part, I know the documentlinkcommandmatcher is not tested, but I don't want to block getting resolve into place.  If we find bugs with that matcher, we can fix it then.

mholo65:
Cool! Will try to take this for a spin tomorrow! Great work!

david-driscoll:
Any objections if I merge this?

tintoy:
Sounds fine to me :)

PR #78, opened 2018-02-25
Ensure shutdown handler runs asynchronously

tintoy:
The shutdown handler needs to run asynchronously because otherwise the server may deadlock on shutdown (server disposal disposes the `ProcessScheduler`, which attempts to join its own dispatcher thread; if we're not asynchronous, then the thread that's currently running is blocking that dispatcher thread).


tintoy:
Given the way that the dispatcher thread currently works (calls the `Handle()` method which may block before returning its `Task` if it's not really async), probably all of our handlers should be genuinely async, but certainly ones with reentrant behaviour such as the shutdown one need to be :)

Issue #83, opened 2018-03-19
Completion handler cannot get completion context sent by client


ghost:
I'm working on a language server and using Vscode as client. Vscode client supports completion context, but ICompletionHandler provides only a TextDocumentPositionParams parameter. When I try to cast it to a CompletionParams, I get an InvalidCastException. Is there no way I can extract the completion context ?

david-driscoll:
That's definitely a bug, I'll fix that with the next push.

PR #84, opened 2018-05-07
Spec/updates

### LINKED ISSUE #82: SignatureHelp.ActiveParameter and SignatureHelp.ActiveSignature should be nullable
As far as I can tell from [the docs](https://microsoft.github.io/language-server-protocol/specification#textDocument_signatureHelp), `ActiveParameter` and `ActiveSignature` should be of type `int?`.

These properties already have the `[Optional]` attribute, but it looks like the nullable suffix was missed off.

https://github.com/OmniSharp/csharp-language-server-protocol/blob/20b3af2f1a9d37fccc7de4a1b0cc545a0e9f5f5e/src/Protocol/Models/SignatureHelp.cs#L23

david-driscoll:
Fixes #82, #83

Brings up to date with the latest lsp spec.


tintoy:
Do we support multiple workspaces yet? That seems like a pretty big chunk of work and the API changes may be tricky to do in a backward-compatible opt-in manner...

david-driscoll:
@tintoy this adds initial support for the workspaces models.  Right now it will only advertise workspaces if you implement a workspace handler.

src/Server/LanguageServer.cs
```diff
@@ -209,9 +212,24 @@ public IDisposable AddHandlerMatcher(IHandlerMatcher handlerMatcher)
...
                 RenameProvider = ccp.HasStaticHandler(textDocumentCapabilities.Rename),
                 SignatureHelpProvider = ccp.GetStaticOptions(textDocumentCapabilities.SignatureHelp).Get<ISignatureHelpOptions, SignatureHelpOptions>(SignatureHelpOptions.Of),
-                WorkspaceSymbolProvider = ccp.HasStaticHandler(workspaceCapabilities.Symbol)
+                WorkspaceSymbolProvider = ccp.HasStaticHandler(workspaceCapabilities.Symbol),
+                ImplementationProvider = ccp.GetStaticOptions(textDocumentCapabilities.Implementation).Get<IImplementationOptions, ImplementationOptions>(ImplementationOptions.Of),
+                TypeDefinitionProvider = ccp.GetStaticOptions(textDocumentCapabilities.TypeDefinition).Get<ITypeDefinitionOptions, TypeDefinitionOptions>(TypeDefinitionOptions.Of),
+                ColorProvider = ccp.GetStaticOptions(textDocumentCapabilities.ColorProvider).Get<IColorOptions, StaticColorOptions>(ColorOptions.Of),
             };
 
+            if (_collection.ContainsHandler(typeof(IDidChangeWorkspaceFoldersHandler)))
```
david-driscoll: @tintoy we can make changes here to help determine if we should support workspace folders, however it's up to the implementor if they support them or not.

david-driscoll:
Going to merge for now... to focus on the other related pull request.  want to try and get both in this weekend, so I can update OmniSharp with it.

PR #85, opened 2018-05-09
Adds a new LanguageServer Builder API

david-driscoll:
This is a large... breaking change, but all existing tests pass.

* Replaces custom mediator framework with `MediatR`
* Adds support for dependency injection internally.
* Added some test helpers to make testing a little easier.
* Bumped target framework to `netstandard2.0`


tintoy:
So is DI internal-only, or can consumers add their own stuff to the container?

Am just thinking it would be useful to register handlers that way if they have dependencies...

david-driscoll:
That's on the list next, first stab was getting things wired up.  Next step is how best to add handlers.

tintoy:
If you're open to a little bike-shedding, I might have some ideas around that (e.g. both Autofac's registration API and the ASP.NET. NET WebAPI controller registration behaviour have ergonomics it might be worth exploring...)

david-driscoll:
Go right ahead!  🚲 🏠 away.

I just added (an untested part) that does pull the handlers from the container as well.

tintoy:
I think as long as there is a way to manually control which handlers are registered (and when) then it'll be fine. I'll write up some thoughts this arvo and then post them here :)

david-driscoll:
Those methods make sense.  Just wrapping a little bow around some of the pieces.

david-driscoll:
Rebased.

sample/SampleServer/Program.cs
```diff
@@ -28,10 +28,9 @@ static void Main(string[] args)
                     .WithLoggerFactory(new LoggerFactory())
                     .AddDefaultLoggingProvider()
                     .WithMinimumLogLevel(LogLevel.Trace)
+                    .WithHandler<TextDocumentHandler>()
```
tintoy: BTW, looks to me like this overload will auto-register any JSON-RPC interfaces implemented by the handler type. Is that correct?

david-driscoll: yeah, the handler collection will create a descriptor per `IJsonRpcHandler` interface it finds in the implementing class.

david-driscoll:
Going to merge this for now, and test one of the pre-release packages against omnisharp.

tintoy:
I'll look into adding similar support for the client this weekend :)

PR #90, opened 2018-06-29
Fix post-initialize request handling.

NTaylorMullen:
- Prior to this when a non-initialize request came the underlying connection would spin up a brand new instance of LspReceiver which never got initialized. Therefore you'd repeatedly get server is not initialized error messages on any request.


src/Server/LanguageServer.cs
```diff
@@ -104,7 +104,7 @@ public static Task<ILanguageServer> From(Action<LanguageServerOptions> optionsAc
             services.AddSingleton(_serializer);
             services.AddSingleton<OmniSharp.Extensions.JsonRpc.ISerializer>(_serializer);
             services.AddSingleton(requestProcessIdentifier);
-            services.AddSingleton(_reciever);
+            services.AddSingleton<IReceiver>(_reciever);
```
david-driscoll: probably need it registered for both `IReciever` and `ILspReciever`.

NTaylorMullen: Didn't see that itw as a different type in this ctor. Ya you're right. Will do that.

NTaylorMullen: Done.

david-driscoll:
The build error looks legit (missing namespace).

NTaylorMullen:
Just got back from vaca. Thanks for fixing my mishap @david-driscoll 👍

PR #100, opened 2018-08-20
[WIP] Start implementing code folding provider

NickAcPT:
Implements the `textDocument/foldingRange` request.

Title contains `WIP` because I'm yet to be able to make vscode work with it (haven't tried other editors).

Thank you all for this amazing implementation of the protocol.

### DIFF
--- a/LanguageServer/Protocol/Document/Server/IFoldingRangeHandler.cs
+++ b/LanguageServer/Protocol/Document/Server/IFoldingRangeHandler.cs
@@ -1,0 +1,11 @@
+using OmniSharp.Extensions.JsonRpc;
+using OmniSharp.Extensions.LanguageServer.Protocol.Client.Capabilities;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+
+// ReSharper disable once CheckNamespace
+namespace OmniSharp.Extensions.LanguageServer.Protocol.Server
+{
+    using static DocumentNames;
+    [Serial, Method(FoldingRange)]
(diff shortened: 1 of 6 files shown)


LanguageServer/Protocol/Document/Server/IFoldingRangeHandler.cs
```diff
@@ -0,0 +1,11 @@
+using OmniSharp.Extensions.JsonRpc;
+using OmniSharp.Extensions.LanguageServer.Protocol.Client.Capabilities;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+
+// ReSharper disable once CheckNamespace
+namespace OmniSharp.Extensions.LanguageServer.Protocol.Server
+{
+    using static DocumentNames;
+    [Serial, Method(FoldingRange)]
```
david-driscoll: Is `Serial` required?  Generally for "answering questions" `Parallel` behaves nicer because we can return results back to the editor faster.

david-driscoll:
@NickAcPT thanks for the contribution, it looks like there may be a few types missing, however I'm going to merge this in as we're due for a spec update anyway, and I'll get to that tomorrow.

Also if you have any feedback or issues you'd like to see improved let me know!

LanguageServer/Protocol/Document/Server/IFoldingRangeHandler.cs
NickAcPT: Sorry for the late answer. While implementing these classes, I've had a look at other classes and tried to follow what they do..

PR #102, opened 2018-08-28
Bug/gh #98

### LINKED ISSUE #98: Should send jsonrpc instead of protocolVersion in messages
Hello,

I finally figured out why this wasn't working with lsp-mode, and it has to do with the fact that lsp-mode is looking specifically for the jsonrpc key in the message, which this seems to instead call protocolVersion. In looking at the spec, it seems that jsonrpc is the correct key, at least it is what's shown in the examples. Please let me know if I am missing something.

Thanks!

david-driscoll:
Fixes #98 

Instead of serializing out `{"jsonrpc": "2.0", "result": 19, "id": 1}` as per https://www.jsonrpc.org/specification we've been serializing `{"protcolVersion": "2.0", "result": 19, "id": 1}`.  Talk about a WTF moment...

image

### DIFF
--- a/src/Client/Protocol/ServerMessage.cs
+++ b/src/Client/Protocol/ServerMessage.cs
@@ -14,6 +14,7 @@ namespace OmniSharp.Extensions.LanguageServer.Client.Protocol
         /// <summary>
         ///     The JSON-RPC protocol version.
         /// </summary>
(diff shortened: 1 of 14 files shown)


david-driscoll:
At least there tests were correctly wrong... 😄

PR #118, opened 2019-01-11
Adds a test for running Language Server.  Fixes a small issue with in…

david-driscoll:
…itialization

### DIFF
--- a/src/Server/LspRequestRouter.cs
+++ b/src/Server/LspRequestRouter.cs
@@ -104,7 +104,7 @@ namespace OmniSharp.Extensions.LanguageServer.Server
                         else
                         {
                             _logger.LogDebug("Converting params for Notification {Method} to {Type}", notification.Method, descriptor.Params.FullName);
-                            var @params = notification.Params.ToObject(descriptor.Params, _serializer.JsonSerializer);
+                            var @params = (notification.Params ?? new JObject()).ToObject(descriptor.Params, _serializer.JsonSerializer);
 
                             await MediatRHandlers.HandleNotification(mediator, descriptor, @params ?? EmptyRequest.Instance, CancellationToken.None);
                         }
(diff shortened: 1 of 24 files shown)


test/Lsp.Tests/LanguageServerTests.cs
```diff
@@ -0,0 +1,54 @@
...
+                //.WithHandler(handler)
+                .WithInput(process.ClientOutputStream)
+                .WithOutput(process.ClientInputStream)
+                .WithLoggerFactory(LoggerFactory)
+                .AddDefaultLoggingProvider()
+                .WithMinimumLogLevel(LogLevel.Trace),
+                cts.Token
+            );
+
+            await Task.WhenAll(
```
tintoy: Neat :)

tintoy:
Awesome - good to see we're using our own client! I meant to get around to this a while back but keep getting dragged off onto other things, so thanks for getting it done :)

src/Server/LspRequestRouter.cs
```diff
@@ -104,7 +104,7 @@ public async Task RouteNotification(IHandlerDescriptor descriptor, Notification
                         else
                         {
                             _logger.LogDebug("Converting params for Notification {Method} to {Type}", notification.Method, descriptor.Params.FullName);
-                            var @params = notification.Params.ToObject(descriptor.Params, _serializer.JsonSerializer);
+                            var @params = (notification.Params ?? new JObject()).ToObject(descriptor.Params, _serializer.JsonSerializer);
```
david-driscoll: @mholo65 this will probably fix the initialization issue, as it allows a null params though now that the IInitializedHandler has a type (it didn't used to).

mholo65: @david-driscoll, I’m afraid this isn’t related to the issue I’m seeing. The issue I’m having is a chicken-egg situation. I can’t create the handler until during initialize, because O# needs to know the target directory in order to createt the OmnisharpWorkspace. But in initialize, there’s no possibility to register the handlers because a) the factory method haven’t yet returned the LanguageServer instance and b) the IServiceCollection in Options used to instantiate the language server have already been built. In previous versions of this lib it was easier since we had access to the instance of the server inside our init delegate.

PR #119, opened 2019-01-13
Pass ILanguageServer instance in InitializeDelegate

mholo65:
- Makes it possible to register new handlers during initialization phase.


mholo65:
@david-driscoll this should solve the problems I'm facing when upgrading in OmniSharp-Roslyn.

If you like to avoid a breaking change, I can rework this PR.

PR #126, opened 2019-03-14
Upgrade/latest lsp

david-driscoll:
This updates to the latest lsp version, and also fixes a couple issues.

cc @mholo65 @tintoy @gep13


src/Protocol/Models/DidChangeConfigurationParams.cs
```diff
@@ -10,6 +11,6 @@ public class DidChangeConfigurationParams : IRequest
         /// <summary>
         ///  The actual changed settings
         /// </summary>
-        public IDictionary<string, BooleanNumberString> Settings { get; set; } = new Dictionary<string, BooleanNumberString>();
+        public JToken Settings { get; set; }
```
gep13: Woot! 😄

david-driscoll: I feel like it changed at somepoint, I just never noticed it.  Keeping a reference spec in sync is hard...

gep13: I bet!  You are doing an amazing job here working on this, and I very much appreciate your hard work!

PR #127, opened 2019-03-16
Add missing FailureHandlingKind enum member.

NickAcPT:
This adds the missing TextOnlyTransactional enum value to FailureHandlingKind.cs.
Without this value, my LSP would crash with the 1.33.0 vscode insiders version.

cc @david-driscoll


NickAcPT:
Closing due to the merging of #129.

PR #128, opened 2019-03-17
Added delegate handlers for all base interfaces

david-driscoll:
@tintoy We should now have parity between the client and server in terms of requests / notifications for both sides.  We should look at the client to figure out how best to pull those in.


tintoy:
Ok - will look into surfacing those APIs on the client next week :)

tintoy:
Nice work on propagating cancellation tokens :)

david-driscoll:
@NTaylorMullen This will involve some (hopefully minor) breaking changes, I'll make the next version number reflect this.  Let me know if you have any concerns, it'll be a few days before I get back to this.

NTaylorMullen:
All good. WIll consume once they're in 😄

PR #133, opened 2019-04-03
Support Client capabilities implementing ConnectedCapability<> multiple times

### LINKED ISSUE #132: LSP Server not working in Visual Studio
I have been setting up an LSP implementation for Chocolatey nuspec files in VSCode, and this has been working well.  I am now trying to take the same Language Server implementation and run it in Visual Studio.  When I try this, I get the following error returned:

image

The code that I am using the initialise the client within Visual Studio is the following:

```
    [ContentType("nuspec")]
    [Export(typeof(ILanguageClient))]
    public class ChocolateyLanguageClient : ILanguageClient
    {
        public string Name => "Chocolatey Language Server Client";

        public IEnumerable<string> ConfigurationSections => null;

        public object InitializationOptions => null;

        public IEnumerable<string> FilesToWatch => null;

        public event AsyncEventHandler<EventArgs> StartAsync;
        public event AsyncEventHandler<EventArgs> StopAsync;

        public async Task<Connection> ActivateAsync(CancellationToken token)
        {
            await Task.Yield();

            var info = new ProcessStartInfo();
            info.FileName = "dotnet";
            info.Arguments = "C:\\github_local\\gep13\\chocolatey-vs\\lib\\ChocolateyLanguageServer\\Chocolatey.Language.Server.dll";
            info.RedirectStandardInput = true;
            info.RedirectStandardOutput = true;
            info.UseShellExecute = false;
            info.CreateNoWindow = true;

            Process process = new Process();
            process.StartInfo = info;

            if (process.Start())
            {
                return new Connection(process.StandardOutput.BaseStream, process.StandardInput.BaseStream);
            }

            return null;
        }

        public async Task OnLoadedAsync()
        {
            await StartAsync?.InvokeAsync(this, EventArgs.Empty);
        }

        public Task OnServerInitializedAsync()
        {
            return Task.CompletedTask;
        }

        public Task OnServerInitializeFailedAsync(Exception e)
        {
            return Task.CompletedTask;
        }
    }
```

Is there additional configuration that needs to be provided, or is this a bug that needs to be addressed.  Let me know if you need any further information.

mholo65:
- Fixes #132

### DIFF
--- a/test/Lsp.Tests/ClientCapabilityProviderTests.cs
+++ b/test/Lsp.Tests/ClientCapabilityProviderTests.cs
@@ -18,24 +18,9 @@ namespace Lsp.Tests
 {
     public class ClientCapabilityProviderTests
     {
-        private static readonly Type[] Capabilities = {
-            typeof(CompletionCapability),
-            typeof(HoverCapability),
-            typeof(SignatureHelpCapability),
-            typeof(ReferencesCapability),
-            typeof(DocumentHighlightCapability),
-            typeof(DocumentSymbolCapability),
-            typeof(DocumentFormattingCapability),
-            typeof(DocumentRangeFormattingCapability),
-            typeof(DocumentOnTypeFormattingCapability),
-            typeof(DefinitionCapability),
-            typeof(CodeActionCapability),
-            typeof(CodeLensCapability),
-            typeof(DocumentLinkCapability),
-            typeof(RenameCapability),
-            typeof(WorkspaceSymbolCapability),
-            typeof(ExecuteCommandCapability),
-        };
+        private static readonly Type[] Capabilities = typeof(ClientCapabilities).Assembly.GetTypes()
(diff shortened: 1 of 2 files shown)


test/Lsp.Tests/ClientCapabilityProviderTests.cs
```diff
@@ -18,24 +18,9 @@ namespace Lsp.Tests
...
-            typeof(DocumentOnTypeFormattingCapability),
-            typeof(DefinitionCapability),
-            typeof(CodeActionCapability),
-            typeof(CodeLensCapability),
-            typeof(DocumentLinkCapability),
-            typeof(RenameCapability),
-            typeof(WorkspaceSymbolCapability),
-            typeof(ExecuteCommandCapability),
-        };
+        private static readonly Type[] Capabilities = typeof(ClientCapabilities).Assembly.GetTypes()
```
mholo65: I thought it was better to load these via reflection, as the tricky ones where missing in the static list (e.g. SynchronizationCapability).

Issue #134, opened 2019-04-04
[LSP] textDocument/documentColor RGB values should be floating point


jeremybeier:
## Issue
The RGB values should be of type double, not int.
> csharp-language-server-protocol/src/Protocol/Models/Color.cs 

### According to the [lsp inspector](https://microsoft.github.io/language-server-protocol/inspector/):
![image](https://user-images.githubusercontent.com/22215375/55558181-cdb4eb00-56eb-11e9-81b2-b329a3468409.png)

mholo65:
Fixed in #135

PR #135, opened 2019-04-04
RGB values for colors are now represented using floats

theBadT:
Proposal for #134


tintoy:
`float` or `double`?

theBadT:
> `float` or `double`?

float

david-driscoll:
sounds good to me!

Issue #136, opened 2019-04-04
Wrong IRegistration definition


gep13:
Following on from the fix that was provided here:

https://github.com/OmniSharp/csharp-language-server-protocol/issues/132

When trying to run the Chocolatey Language Server again in Visual Studio, I was met with another error:

`Error converting value {null} to type 'System.Boolean'. Path 'capabilities.codeActionProvider'.`

Full stack trace here:

<details>
  at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.EnsureType(JsonReader reader, Object value, CultureInfo culture, JsonContract contract, Type targetType)
at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
  at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.SetPropertyValue(JsonProperty property, JsonConverter propertyConverter, JsonContainerContract containerContract, JsonProperty containerProperty, JsonReader reader, Object target)
  at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)
  at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
  at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
  at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.SetPropertyValue(JsonProperty property, JsonConverter propertyConverter, JsonContainerContract containerContract, JsonProperty containerProperty, JsonReader reader, Object target)
  at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)
[...]

PR #139, opened 2019-05-21
Respect client capabilities when choosing codeActionProvider return type

tgjones:
Possible fix for #136, but I'm new to this project so I've quite possibly misunderstood what needs to be done.

### DIFF
--- a/src/Server/LanguageServer.cs
+++ b/src/Server/LanguageServer.cs
@@ -296,7 +296,9 @@ namespace OmniSharp.Extensions.LanguageServer.Server
 
             var serverCapabilities = new ServerCapabilities()
             {
-                CodeActionProvider = ccp.GetStaticOptions(textDocumentCapabilities.CodeAction).Get<ICodeActionOptions, CodeActionOptions>(CodeActionOptions.Of),
+                CodeActionProvider = textDocumentCapabilities.CodeAction.IsSupported && textDocumentCapabilities.CodeAction.Value.CodeActionLiteralSupport != null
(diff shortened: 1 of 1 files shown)


mholo65:
@tgjones thanks for your PR... Some time ago, I went trying to fix this issue in another way (the null serialization), see #140.

However, this includes checks for respecting client request, which is a good thing. I "think" we should create a factory method in `ServerCapabilities` for constructing it, as mentioned in comments:
https://github.com/OmniSharp/csharp-language-server-protocol/blob/62cda45b1eff416c6935b4d9a58f0486f9f34184/src/Protocol/Server/Capabilities/ServerCapabilities.cs#L13-L15

david-driscoll:
Going to close this for the moment, as the fix doesn't seem appropriate, but I'll give some context (because the code isn't well documented on this!)

`GetStaticOptions` does a check for `IsSupported`.  It also does a check to make sure that there is an object value that came from the client (the editor).  And it checks for `DynamicRegistration` .  VsCode does something really interesting here when you deal with Dynamic Registration.

If we report that we support code lens at this point, vscode will statically bind code lens to whatever selectors were provided with the extension.  We then process any handlers that support dynamic registration and after initialization we issue all the registration requests.  This then causes a double registration to happen.  So we actually get bound to code lens once for the provided selectors (with the extension) as well as any extensions that we provided through dynamic registration.  This is why the server will in some cases report that it doesn't support something, because we will after the fact bind dynamically to the client.

PR #148, opened 2019-07-13
Updated support for capabilities to be more generous to clients that …

david-driscoll:
…do not report their support via capabilities


david-driscoll:
see #149

david-driscoll:
fixes #149

Issue #162, opened 2019-08-07
textDocumentSync coming back as 0/None


tom-bowles:
When using OmniSharp LSP from emacs lsp-mode, the initialize handshake is as shown at the bottom. The server response contains `"textDocumentSync": 0`, so lsp-mode sends no textDocument/didChange notifications.

The relevant code in `LanguageServer.cs` is also shown below. The check for `ccp.HasStaticHandler(textDocumentCapabilities.Synchronization)` returns false because although there is a `TextDocumentSyncHandler` registered (specifically `OmnisharpTextDocumentSyncHandler`), `TextDocumentSyncHandler` doesn't provide `IWillSaveTextDocumentHandler` or `IWillSaveWaitUntilTextDocumentHandler`, both of which are linked to `SynchronizationCapability`.

Removing the check causes the server to return a textDocumentSync options block, but the `change` item is set to 1 (full) rather than 2 (incremental). This appears to be because there are two OmnisharpTextDocumentSyncHandlers registered, one for cake and one for cs/csx, and the cake one doesn't handle incremental. Not sure if that should be a separate issue for omnisharp-roslyn, but it means you can't send incremental changes for cs/csx files because the Min() call in the code below means it picks up kind 1.

```
            if (ccp.HasStaticHandler(textDocumentCapabilities.Synchronization))
            {
                var textDocumentSyncKind = _collection.ContainsHandler(typeof(IDidChangeTextDocumentHandler))
                    ? _collection
                        .Select(x => x.Handler)
                        .OfType<IDidChangeTextDocumentHandler>()
                        .Where(x => x.GetRegistrationOptions()?.SyncKind != TextDocumentSyncKind.None)
                        .Min(z => z.GetRegistrationOptions()?.SyncKind)
                    : TextDocumentSyncKind.None;

                if (_clientVersion == ClientVersion.Lsp2)
                {
                    serverCapabilities.TextDocumentSync = textDocumentSyncKind;
                }
                else
                {
                    serverCapabilities.TextDocumentSync = new TextDocumentSyncOptions()
                    {
                        Change = textDocumentSyncKind ?? TextDocumentSyncKind.None,
                        OpenClose = _collection.ContainsHandler(typeof(IDidOpenTextDocumentHandler)) || _collection.ContainsHandler(typeof(IDidCloseTextDocumentHandler)),
                        Save = _collection.ContainsHandler(typeof(IDidSaveTextDocumentHandler)) ?
[...]

PR #163, opened 2019-08-07
Changes for #162 correct how text document sync settings are configured

david-driscoll:


david-driscoll:
This should hopefully fix some of the issues seen in #162

src/Protocol/Server/Capabilities/TextDocumentSyncOptions.cs
```diff
@@ -38,17 +39,24 @@ public class TextDocumentSyncOptions : ITextDocumentSyncOptions
...
-            return new TextDocumentSyncOptions() {
-                OpenClose = options.Any(z => z.OpenClose),
-                Change = options
+            var change = TextDocumentSyncKind.None;
+            if (options.Any())
+            {
+                change = options
                         .Where(x => x.Change != TextDocumentSyncKind.None)
-                        .Min(z => z.Change),
+                        .Min(z => z.Change);
```
mholo65: As per comment in https://github.com/OmniSharp/csharp-language-server-protocol/issues/162#issuecomment-519433579 this seems to throw because sequence doesn't contain any elements. Maybe add the same predicate `x => x.Change != TextDocumentSyncKind.None` also to the `if (options.Any())` check above?

david-driscoll: sounds like a good change...

david-driscoll:
@mholo65 updated 🙃

PR #168, opened 2019-09-11
JsonRpcServer now disposable

TylerLeonhardt:
`JsonRpcServer.Dispose()` now works. It use to throw an exception deep from within `InputHandler` and `OutputHandler`.

Not sure if this needs any test changes but I think the tests need fixing anyway from previous PRs.


TylerLeonhardt:
This PR could probably go into master already? Not 100% sure.

PR #171, opened 2019-09-23
Initial handling of errors

TylerLeonhardt:
This PR includes the initial ability to handle errors via throwing a `HandledErrorException`.

It also fixes up the `DapRpcErrorConverter` to have the right schema.


TylerLeonhardt:
Actually it looks as if this code path doesn't even work. Any exception I throw ends up in this catch:

https://github.com/OmniSharp/csharp-language-server-protocol/blob/dap/src/JsonRpc/RequestRouterBase.cs#L161-L165

not sure if that's by design...

Issue #175, opened 2019-09-27
DiagnosticRelatedInformation Location should be an Location object


drsoran:
Currently the `DiagnosticRelatedInformation.Location` property is a `string`. I think this is not correct in sense of Language Server Protocol. I think they expect `Location` to be an `Location` object with URI and Range.

I tried to pass a string but VS Code then displays related messages as:

**undefined(undefined, undefined): message**

drsoran:
#176

david-driscoll:
@drsoran I think you're correct.  Thanks for the find!

PR #176, opened 2019-09-28
Changes for #175 to return a Location object for diagnostic related information

drsoran:
This will change the return of the DiagnosticRelatedInformation.Location property from string to Location.


david-driscoll:
Fixes #175

david-driscoll:
Thanks for the help @drsoran !

PR #177, opened 2019-10-01
Do not embed MediatR

mholo65:
MediatR is strong-name signed nowadays, no need to embed and strong-name signed anymore.

// cc @NTaylorMullen and @TylerLeonhardt would this break for any of you? It's basically a couple of more DLLs to ship with the extension, but nothing more.


TylerLeonhardt:
Yes it would require a small change on my part but thanks for the heads up!

TylerLeonhardt:
Let me know when I should expect my stuff to start breaking!

PR #180, opened 2019-10-04
DO NOT MERGE: Work needed to get logging working

TylerLeonhardt:
This was the change I needed in PowerShell Editor Services in order to get logging to a file AND to the Lsp working.

The change I made in PSES as reference is here (there's some commentary in that code change): https://github.com/PowerShell/PowerShellEditorServices/pull/1058


TylerLeonhardt:
I don't think we need this anymore

PR #182, opened 2019-10-06
Refactor logging  +semver:minor

mholo65:
- Expose ILoggingBuilder instead of accepting a ILoggerFactory

Some open questions. How to maintain backwards compatibility, or should we?

### DIFF
--- a/test/Lsp.Tests/LanguageServerTests.cs
+++ b/test/Lsp.Tests/LanguageServerTests.cs
@@ -37,10 +37,10 @@ namespace Lsp.Tests
             var serverStart = LanguageServer.From(x => x
                 //.WithHandler(handler)
                 .WithInput(process.ClientOutputStream)
-                .WithOutput(process.ClientInputStream)
-                .WithLoggerFactory(LoggerFactory)
-                .AddDefaultLoggingProvider()
-                .WithMinimumLogLevel(LogLevel.Trace),
+                .WithOutput(process.ClientInputStream),
+                //.WithLoggerFactory(LoggerFactory)
+                //.AddDefaultLoggingProvider()
+                //.WithMinimumLogLevel(LogLevel.Trace),
                 cts.Token
             );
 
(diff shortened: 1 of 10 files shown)


mholo65:
@david-driscoll is it worth it to work around and try to keep the old API with `ILoggingFactory` in order to don't break any existing users? We're still `0.x` so might as well break 😄 Need to fixup the tests however, and probably fix the minimum loglevel also once for all 🛩

test/Lsp.Tests/LanguageServerTests.cs
```diff
@@ -37,10 +37,10 @@ public async Task Works_With_IWorkspaceSymbolsHandler()
...
                 //.WithHandler(handler)
                 .WithInput(process.ClientOutputStream)
-                .WithOutput(process.ClientInputStream)
-                .WithLoggerFactory(LoggerFactory)
-                .AddDefaultLoggingProvider()
-                .WithMinimumLogLevel(LogLevel.Trace),
+                .WithOutput(process.ClientInputStream),
+                //.WithLoggerFactory(LoggerFactory)
+                //.AddDefaultLoggingProvider()
+                //.WithMinimumLogLevel(LogLevel.Trace),
```
david-driscoll: Can probably safely delete these lines

test/Lsp.Tests/LanguageServerTests.cs
```diff
@@ -63,9 +63,9 @@ public async Task GH141_CrashesWithEmptyInitializeParams()
             var server = LanguageServer.PreInit(x => x
                 .WithInput(process.ClientOutputStream)
                 .WithOutput(process.ClientInputStream)
-                .WithLoggerFactory(LoggerFactory)
-                .AddDefaultLoggingProvider()
-                .WithMinimumLogLevel(LogLevel.Trace)
+                //.WithLoggerFactory(LoggerFactory)
```
david-driscoll: Can probably safely delete these lines

PR #183, opened 2019-10-08
Update cake bootstrap scripts

TylerLeonhardt:
Note the changes to the bootstrap ps1 and sh files come directly from https://github.com/cake-build/resources/tree/master

A small change to the build.cake was needed to set the default target to `Default`.

This means that `build.ps1` also working in PowerShell on non-Windows platforms 😃

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,8 @@
 #load "nuget:?package=Rocket.Surgery.Cake.Library&version=0.9.3";
 
+// Default the target to "Default" if one was not specified.
+var target = Argument("target", "Default");
+
 Task("Submodules")
     .Does(() => {
         StartProcess("git", "submodule update --init --recursive");
@@ -25,4 +28,4 @@ Task("Default")
     .IsDependentOn("Embed MediatR")
     .IsDependentOn("dotnetcore");
 
-RunTarget(Target);
+RunTarget(target);
--- a/build.ps1
+++ b/build.ps1
@@ -21,34 +21,51 @@ The build script target to run.
 The build configuration to use.
 .PARAMETER Verbosity
 Specifies the amount of information to be displayed.
-.PARAMETER WhatIf
-Performs a dry run of the build script.
-No tasks will be executed.
+.PARAMETER ShowDescription
+Shows description about tasks.
+.PARAMETER DryRun
+Performs a dry run.
 .PARAMETER SkipToolPackageRestore
 Skips restoring of packages.
 .PARAMETER ScriptArgs
 Remaining arguments are added here.
 
 .LINK
-http://cakebuild.net
+https://cakebuild.net
 
 #>
 
 [CmdletBinding()]
 Param(
-    [parameter(position=0)]
-    [string]$Target = "Default",
     [string]$Script = "build.cake",
-    [ValidateSet("Release", "Debug")]
-    [string]$Configuration = "Release",
+    [string]$Target,
+    [string]$Configuration,
     [ValidateSet("Quiet", "Minimal", "Normal", "Verbose", "Diagnostic")]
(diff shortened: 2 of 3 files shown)


build.ps1
```diff
@@ -21,35 +21,52 @@ The build script target to run.
...
 .LINK
-http://cakebuild.net
+https://cakebuild.net
 
 #>
 
 [CmdletBinding()]
 Param(
-    [parameter(position=0)]
-    [string]$Target = "Default",
```
david-driscoll: This default was intentional.

build.ps1
```diff
@@ -21,35 +21,52 @@ The build script target to run.
...
     [string]$Script = "build.cake",
-    [ValidateSet("Release", "Debug")]
-    [string]$Configuration = "Release",
+    [string]$Target,
+    [string]$Configuration,
     [ValidateSet("Quiet", "Minimal", "Normal", "Verbose", "Diagnostic")]
-    [string]$Verbosity = "Verbose",
```
david-driscoll: Verbosity was intentional but it doesn't have to be the default.

build.ps1
```diff
@@ -21,35 +21,52 @@ The build script target to run.
...
 
 #>
 
 [CmdletBinding()]
 Param(
-    [parameter(position=0)]
-    [string]$Target = "Default",
     [string]$Script = "build.cake",
-    [ValidateSet("Release", "Debug")]
-    [string]$Configuration = "Release",
```
david-driscoll: Default on configuration was intentional.

build.sh
```diff
@@ -24,24 +28,14 @@ fi
 
 # Define default arguments.
 SCRIPT="build.cake"
-TARGET="Default"
```
david-driscoll: These defaults were intentional

build.cake
```diff
@@ -25,4 +28,4 @@ Task("Default")
     .IsDependentOn("Embed MediatR")
     .IsDependentOn("dotnetcore");
 
-RunTarget(Target);
```
david-driscoll: This is defined in the other package, and is picked up automatically.  The reason this stopped working correctly is due to the breaking changes to the bootstrapper scripts.

TylerLeonhardt:
@david-driscoll, @gep13 was saying in chat that it's a better practice to handle the defaults in the build.cake that way the bootstrap scripts can be identical to what they are in the cake repo.

Would you like me to update the build.cake with the other defaults you called out or modify the bootstrapping script?

david-driscoll:
It has a default here: https://github.com/RocketSurgeonsGuild/Cake/blob/d2466e190736f128f8465adfe4cc6a73eebc5b0c/src/Cake/LocalCakeAliases.cs#L18

It's obscure, but it is defined.  Why it's not defaulted to `Default` is a bug/feature, lol.

If you change the default task from `Default` to `Build` and then the target definition can go away.

For the bootstrapping script I've always made it so that I can run `.\build.ps1 <Target>` because it really bothers me vs other build tools that you have to `.\build.ps1 -Target <Target>`.  Just less characters to type.  I thought I had changed build.sh to that as well, but I don't see it here.

If the only thing that survives is the `[parameter(position=0)]` I'm happy.

TylerLeonhardt:
I think I'm going to abandon this. It sounds like the csharp-lsp build scripts are a bit too much of a fork that I can deal with atm and I don't want to risk breaking them.

Also I have more travel coming up so the likelihood of being able to get this done in a timely manor is slim...

If someone else wants to pick it up, go for it!

Perhaps this is a good opportunity to give feedback to the cake folks as to how the bootstrap script can be improved. I like the addition of a `position=0` for `Target`.

PR #187, opened 2019-10-14
Adds ILoggingBuilder extension for LanguageServer logging.

mholo65:
replaces `AddDefaultLoggingProvider()` with `.AddLanguageServer()`.

Allows you to pass in a `LogLevel` into `.AddLanguageServer()` (default is Information)

### DIFF
--- a/sample/SampleServer/Program.cs
+++ b/sample/SampleServer/Program.cs
@@ -19,7 +19,7 @@ namespace SampleServer
 
         static async Task MainAsync(string[] args)
         {
-            Debugger.Launch();
+            //Debugger.Launch();
             //while (!System.Diagnostics.Debugger.IsAttached)
             //{
             //    await Task.Delay(100);
@@ -38,8 +38,8 @@ namespace SampleServer
                     .WithOutput(Console.OpenStandardOutput())
                     .ConfigureLogging(x => x
                         .AddSerilog()
-                        .AddLanguageServer(LogLevel.Critical)
-                        .SetMinimumLevel(LogLevel.Trace))
+                        .AddLanguageServer(LogLevel.Error)
+                        .SetMinimumLevel(LogLevel.Error))
                     .WithHandler<TextDocumentHandler>()
                     .WithHandler<DidChangeWatchedFilesHandler>()
                     .WithHandler<FoldingRangeHandler>()
--- a/sample/SampleServer/Program.cs
+++ b/sample/SampleServer/Program.cs
@@ -19,10 +19,11 @@ namespace SampleServer
 
         static async Task MainAsync(string[] args)
         {
-            // while (!System.Diagnostics.Debugger.IsAttached)
-            // {
+            Debugger.Launch();
+            //while (!System.Diagnostics.Debugger.IsAttached)
+            //{
             //    await Task.Delay(100);
-            // }
+            //}
 
             Log.Logger = new LoggerConfiguration()
               .Enrich.FromLogContext()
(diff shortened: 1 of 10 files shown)


sample/SampleServer/Program.cs
```diff
@@ -35,8 +36,10 @@ static async Task MainAsync(string[] args)
                 options
                     .WithInput(Console.OpenStandardInput())
                     .WithOutput(Console.OpenStandardOutput())
-                    .ConfigureLogging(x => x.AddSerilog())
-                    .AddDefaultLoggingProvider()
+                    .ConfigureLogging(x => x
+                        .AddSerilog()
+                        .AddLanguageServer(LogLevel.Critical)
+                        .SetMinimumLevel(LogLevel.Trace))
```
mholo65: I'd like to avoid having this, but truth is that MEL adds a default filter which works on the factory level, so we won't get our logs to LSP logger if they are < Information (which is default),

src/Server/LanguageServer.cs
```diff
@@ -357,9 +343,14 @@ private async Task Initialize(CancellationToken token)
...
-            if (request.Trace == InitializeTrace.Verbose && MinimumLogLevel >= LogLevel.Information)
+            if (request.Trace == InitializeTrace.Verbose)
             {
-                MinimumLogLevel = LogLevel.Trace;
+                var loggerSettings = _serviceProvider.GetService<LanguageServerLoggerSettings>();
+
+                if (loggerSettings?.MinimumLogLevel >= LogLevel.Information)
+                {
+                    loggerSettings.MinimumLogLevel = LogLevel.Trace;
+                }
```
mholo65: Best would be if we'd also here could notify `IOptionsMonitor<LoggerFilterOptions>` that it has changed. Without a configuration file, we'd probably have to do something like this. https://stackoverflow.com/a/53077453

david-driscoll: We should probably look at a configuration integration.  My thought is something like `interface ILspConfiguration : IConfiguration` and then populate that some defaults, like the logging state, but also wrap up the configuration handlers so that they surface their data through it, along with the refresh tokens.

sample/SampleServer/Program.cs
tintoy: As long as it's configured by the consumer (read: Host process) I think that's fine.

src/Server/LanguageServer.cs
tintoy: There's a better way - I had to do something similar for KubeClient:

https://github.com/tintoy/dotnet-kube-client/tree/develop/src/KubeClient.Extensions.Configuration

tintoy: Specifically:

https://github.com/tintoy/dotnet-kube-client/blob/develop/src/KubeClient.Extensions.Configuration/ConfigMapConfigurationProvider.cs#L155

tintoy: Then:

https://github.com/tintoy/dotnet-kube-client/blob/develop/src/KubeClient.Extensions.Configuration/ConfigMapConfigurationProvider.cs#L191

mholo65: I solved it by registering an own `IOptionsMonitor<LoggerFilterOptions>`

sample/SampleServer/Program.cs
```diff
@@ -35,8 +36,10 @@ static async Task MainAsync(string[] args)
                 options
                     .WithInput(Console.OpenStandardInput())
                     .WithOutput(Console.OpenStandardOutput())
-                    .ConfigureLogging(x => x.AddSerilog())
-                    .AddDefaultLoggingProvider()
+                    .ConfigureLogging(x => x
+                        .AddSerilog()
+                        .AddLanguageServer(LogLevel.Error)
+                        .SetMinimumLevel(LogLevel.Error))
```
TylerLeonhardt: why do we need to pass in the `LogLevel` for both the `AddLanguageServer` and `SetMinimumLevel`?

mholo65: You don't need to. It's an option. One is global, the other one is just for LSP logging.

TylerLeonhardt: @mholo65 I gave this a go:

```
.AddSerilog()
.AddLanguageServer()
.SetMinimumLevel(LogLevel.Error)
```

and the LSP side was only giving Information and Error.

PR #189, opened 2019-10-28
Allow using global log level as minimum log level for LanguageServerLoggerProvider

mholo65:
@TylerLeonhardt is this what you wanted?

### DIFF
--- a/sample/SampleServer/Program.cs
+++ b/sample/SampleServer/Program.cs
@@ -38,8 +38,8 @@ namespace SampleServer
                     .WithOutput(Console.OpenStandardOutput())
                     .ConfigureLogging(x => x
                         .AddSerilog()
-                        .AddLanguageServer(LogLevel.Error)
-                        .SetMinimumLevel(LogLevel.Error))
+                        .AddLanguageServer()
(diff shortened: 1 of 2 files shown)


sample/SampleServer/Program.cs
```diff
@@ -38,8 +38,8 @@ static async Task MainAsync(string[] args)
                     .WithOutput(Console.OpenStandardOutput())
                     .ConfigureLogging(x => x
                         .AddSerilog()
-                        .AddLanguageServer(LogLevel.Error)
-                        .SetMinimumLevel(LogLevel.Error))
+                        .AddLanguageServer()
+                        .SetMinimumLevel(LogLevel.Information))
```
TylerLeonhardt: Maybe worth changing this to like `Debug` or `Trace` since `Information` is the default? Or is something else the default?

mholo65: You are correct, `Information` is the default level.

TylerLeonhardt:
But my comment wasn't addressed...

mholo65:
> But my comment wasn't addressed...

Addressed here https://github.com/OmniSharp/csharp-language-server-protocol/pull/190

PR #196, opened 2019-12-20
Workaround Unicode characters in URIs .NET bug

### LINKED ISSUE #195: Don't rely on Uri.ToString() in Serializer
In the `AbsoluteUriConverter` class, we use `Uri.ToString()`.

https://github.com/OmniSharp/csharp-language-server-protocol/blob/2b06e038e455bb27ee185da9a6672f4cf59dd7b7/src/Protocol/Serialization/Converters/AbsoluteUriConverter.cs#L54

Unfortunately, the `Uri.ToString()` has [bizarre behavior with Uris with escaped non-ASCII characters](https://github.com/dotnet/runtime/issues/1031).

```
> new Uri("/Users/tyleonha/Code/PowerShell/Misc/测试.ps1").ToString()
file:///Users/tyleonha/Code/PowerShell/Misc/%E6%B5%8B%E8%AF%95.ps1/Users/tyleonha/Code/PowerShell/Misc/测试.ps1
```
Notice the path is written twice?

This only seems to happen on non-Windows as adding a drive letter fixes it... but then it url decodes the characters:
```
> new Uri("file:///C:/Users/tyleonha/Code/PowerShell/Misc/%E6%B5%8B%E8%AF%95.ps1").ToString()
file:///C:/Users/tyleonha/Code/PowerShell/Misc/测试.ps1
```
But then again... since VS Code expects the colon to be escaped in the drive letter... we're still left in a bad state:
```
new Uri("file:///C%3A/Users/tyleonha/Code/PowerShell/Misc/%E6%B5%8B%E8%AF%95.ps1").ToString()
file://%2FC%3A%2FUsers%2Ftyleonha%2FCode%2FPowerShell%2FMisc%2F测试.ps1/C%3A/Users/tyleonha/Code/PowerShell/Misc/测试.ps1
```

TylerLeonhardt:
fixes #195 

This adds some logic to better serialize file `Uri`s. It also caught a few places where we weren't using the `AbsoluteUriConverter`.


src/JsonRpc/RequestRouterBase.cs
```diff
@@ -99,7 +99,7 @@ public virtual async Task<ErrorResponse> RouteRequest(TDescriptor descriptor, Re
                     // TODO: Try / catch for Internal Error
                     try
                     {
-                        if (descriptor == default)
+                        if (descriptor.Equals(default))
```
TylerLeonhardt: @mholo65 @david-driscoll for some reason I needed to make this change even though I didn't touch this code path... not sure why...

src/Protocol/Models/BooleanOr.cs
```diff
@@ -15,7 +15,7 @@ public BooleanOr(bool value)
             _bool = value;
         }
 
-        public bool IsValue => this._value != default;
+        public bool IsValue => !this._value.Equals(default);
```
TylerLeonhardt: @mholo65 @david-driscoll same with this...

src/JsonRpc/RequestRouterBase.cs
TylerLeonhardt: I reverted this... but I don't know why it's happening and don't think it's related to my change.

src/Protocol/Models/BooleanOr.cs
TylerLeonhardt: I reverted this... but I don't know why it's happening and don't think it's related to my change.

src/JsonRpc/RequestRouterBase.cs
mholo65: Problem is still on building in Azure DevOps as it's building with .NET SDK 3.1.100... I'll try to fix this asap.

mholo65:
@TylerLeonhardt your changes have been merged! Thank you for your contribution!

PR #199, opened 2020-01-13
Fix TextDocument synchronization for LSP3

NTaylorMullen:
- For LSP3 clients the TextDocumentSynchronization kind is always set to none without this. For the VS LSP client this results in language servers never getting any text document synchronization events.

### DIFF
--- a/src/Server/LanguageServer.cs
+++ b/src/Server/LanguageServer.cs
@@ -458,7 +458,7 @@ namespace OmniSharp.Extensions.LanguageServer.Server
                 {
                     serverCapabilities.TextDocumentSync = new TextDocumentSyncOptions()
                     {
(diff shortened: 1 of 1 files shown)


NTaylorMullen:
I'm not sure if this is the "right" change but it seems odd that for LSP3 scenarios the sync kind is always set to none

NTaylorMullen:
@david-driscoll this seems like a pretty bad one. What're your thoughts? To me it seems like VSCode ignores that "None" entirely

ToddGrun:
I just hit this issue today too. Would love to see this one fixed.

david-driscoll:
So I think the original assumption (as to why this is like this) was related to static vs dynamic handlers and how the protocol handled those.  I think this should be a workable solution to handle the static case.

PR #207, opened 2020-01-31
Added support for ILanguageServerConfiguration

david-driscoll:
This is a wip that will need some more unit tests.

In essence the goal here is to enable support for both `workspace/didChangeConfiguration` and `workspace/configuration` through the MEC `IConfiguration` interface. such that as configuration changes happen on the client, they are reflected correctly in the configuration of the server.  This opens the door to using `IOptionsMonitor` in places as well.

Another part is to allow support for the `scopeUri` portion of `workspace/configuration` returning a disposable interface that is also `IConfiguration` and managing the state for that object as long it as it hasn't been dispoed.  For example keeping the configuration of all the currently opened files is probably something that will be needed.

And finally allowing configuration to just be queried and inspected for different areas of the client.  Basically just to make the work of getting that information a little easier to consume.


david-driscoll:
@NTaylorMullen @mholo65 @tintoy @TylerLeonhardt any feedback you want to give before I push this through? 🌵

TylerLeonhardt:
cc @rjmholt

@david-driscoll this is great - we use our own (very basic) [ConfigurationService](https://github.com/PowerShell/PowerShellEditorServices/blob/master/src/PowerShellEditorServices/Services/Workspace/ConfigurationService.cs) and [Configuration Handler](https://github.com/PowerShell/PowerShellEditorServices/blob/master/src/PowerShellEditorServices/Services/Workspace/Handlers/ConfigurationHandler.cs) today and I'm wondering how we might be able to use these changes.

We [keep track](https://github.com/PowerShell/PowerShellEditorServices/blob/master/src/PowerShellEditorServices/Services/Workspace/LanguageServerSettings.cs#L386-L402) of "powershell" scoped settings and also a couple other generic ones ("files" and "search").

src/Server/ILanguageServerConfiguration.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Gets the current configuration for a given document uri
+        /// This re-uses all the sections from the <see cref="ConfigurationItem"/>s that
+        /// the root configuration uses.
+        ///
+        /// This will watch for changes of the scoped documents and update the configuration.
+        /// </summary>
+        /// <param name="scopeUri"></param>
+        /// <returns></returns>
+        Task<IDisposableConfiguration> GetScopedConfiguration(Uri scopeUri);
```
NTaylorMullen: Hmm, is there value in exposing that this is an IDisposableConfiguration here? Is the LanguageServer author expected to call dispose on it?

david-driscoll: Yeah, the high level idea here is that a language server wants to get configuration information for a specific document (at least from the vscode side anyway).  This would allow them to see any overrides that might have happened for example vscode lets you do `"[csharp]":{}` to override settings for a specific language.

Since there is nothing in the protocol today that lets us ask for settings for a generic language or file type, we instead have to go this route.   The reason it's disposable is because we don't want to have to double information for files that are no longer in scope, as each refresh will update all the active scopes.

NTaylorMullen: I understand that aspect of it but by exposing the IDisposeableness of the scoped configuration I as the Razor langauge server author could call dispose which wouldn't be expected I'd imagine. I'd think you (as O# lib) would want to reserve the right to dispose

src/Server/Configuration/DidChangeConfigurationProvider.cs
```diff
@@ -0,0 +1,174 @@
...
+                    Items = configurationItems
+                })).ToArray();
+
+                foreach (var (scope, settings) in configurationItems.Zip(configurations,
+                    (scope, settings) => (scope, settings)))
+                {
+                    ParseClientConfiguration(settings, scope.Section);
+                }
+
+                OnReload();
```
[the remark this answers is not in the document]
david-driscoll: That's an interesting observation... they will likely get double updated (once from the OnReload() and once from the update call.  It should probably move to the bottom just for consistency.

src/Server/Configuration/DidChangeConfigurationProvider.cs
```diff
@@ -0,0 +1,174 @@
...
+        {
+            // null means we need to re-read the configuration
+            // https://github.com/Microsoft/vscode-languageserver-node/issues/380
+            if (request.Settings == null || request.Settings.Type == JTokenType.Null)
+            {
+                await GetWorkspaceConfiguration();
+                return Unit.Value;
+            }
+
+            ParseClientConfiguration(request.Settings);
```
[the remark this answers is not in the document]
david-driscoll: Do you have examples of a client that gives back xml as configuration (just my knowledge/testing)?  I'll look at adding the service this weekend.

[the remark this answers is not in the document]
NTaylorMullen: So by default VS stores its configurations in XML format however the LSP client for VS doesn't currently support interacting with those XML based configurations. Therefore, for the time being they brought over JSON configurations to act as a replacement of the XML based configurations but all of the new JSON configs do not translate to XML and do not auto-map to the XML configs either. Therefore VS is trying to figure out how it wants its configuration story to roll out so it can work with new language server.

tl;dr; I don't know of any currently; however, it's definitely possible it'll come down in the future.

src/Server/ILanguageServerConfiguration.cs
david-driscoll: The problem is on the LSP side I don't know when you're "done" with that configuration, hence the disposable aspect of it.

It could be in context of a file or it could be in context of a file that is currently being acted upon.  In essence you wanted the configuration for that file, and you take ownership of when you're done. On the library side we're only going to update the configuration for you automagically.

I could be something that is unnecessary and maybe `GetConfiguration` is enough.

NTaylorMullen: Ahh I see

PR #213, opened 2020-02-25
Expose Initialize as a public method on ILanguageServer.

NTaylorMullen:
- For in-proc scenarios the pattern for initializing the language server involves calling PreInit, and then manually calling Initialize. For example that's done here: https://github.com/dotnet/aspnetcore-tooling/pull/1628/files#diff-9327c1670bd2b8bebc131b7dd8e1857bR53
- Memoized the Initialize task to ensure that callign it multiple times only ever initializes once.
- Moved the Initialize method up to ensure it sat next to other public methods.

### DIFF
--- a/src/Server/LanguageServer.cs
+++ b/src/Server/LanguageServer.cs
@@ -52,6 +52,7 @@ namespace OmniSharp.Extensions.LanguageServer.Server
         private readonly CompositeDisposable _disposable = new CompositeDisposable();
         private readonly IServiceProvider _serviceProvider;
         private readonly SupportedCapabilities _supportedCapabilities;
+        private Task _initializingTask;
 
         public static Task<ILanguageServer> From(Action<LanguageServerOptions> optionsAction)
         {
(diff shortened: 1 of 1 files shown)


src/Server/LanguageServer.cs
```diff
@@ -301,38 +302,26 @@ public IDisposable AddTextDocumentIdentifier(params ITextDocumentIdentifier[] ha
...
             }
-        }
 
-        private async Task Initialize(CancellationToken token)
-        {
-            _connection.Open();
             try
             {
-                await _initializeComplete
+                _connection.Open();
```
david-driscoll: If the connection is already opened does this have any effect?

NTaylorMullen: If the connection is already open and you attempt to call open again you'll get a:

image

Because this is effectively what calling open does.

That being said, how would one get access to the `_connection` object to call open twice? I ask because at the top of this method I ensure this `_connection.Open` can't be called more than once.

david-driscoll: I'm slow I didn't see the old connection open call in the original diff.

PR #215, opened 2020-03-08
Fixes #214 by returning LSP conform JSON for prepare rename request

### LINKED ISSUE #175: DiagnosticRelatedInformation Location should be an Location object

### LINKED ISSUE #214: Prepare Rename request handler returns wrong JSON
I think the return value `RangeOrPlaceholderRange` of `textDocument/prepareRename` handler is not serialized correctly. According to LSP it expects:

> result: Range | { range: Range, placeholder: string } | null

But the client gets the full `RangeOrPlaceholderRange` object as JSON that contains a Range as well as { range: Range, placeholder: string }. VS Code, e.g., rejects this object.

I think there is a `RangeOrPlaceholderRangeConverter` missing in the Serializer.

drsoran:
This will add a new converter for RangeOrPlaceholderRange to the Serializer. 

@RangeOrPlaceholderRangeConverter

I let `CanRead` return `false` because the client will never send such an object. I also return an empty `RangeOrPlaceholderRange` in case of `Read`. Not sure, perhaps it`s better to return null or throw?

Fixes #214

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,1 @@
+theme: jekyll-theme-hacker
--- a/_config.yml
+++ b/_config.yml
@@ -1,1 +1,0 @@
-theme: jekyll-theme-hacker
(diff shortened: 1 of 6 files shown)


david-driscoll:
@drsoran thanks!  I'll get this in today.

PR #186, opened 2019-10-13
WIP: Update to the prerelease 3.15 version of spec

david-driscoll:
@mholo65 @TylerLeonhardt 

This is a huge WIP based on the 3.15 spec https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/

PR #216, opened 2020-03-10
Add AllCommitChars to (I)CompletionOptions to add json serialization …

ToddGrun:
…for this during the intialization handshake. Without this, my omnisharp server in VS doesn't receive the AllCommitCharacters for the CompletionOptions.


NTaylorMullen:
@david-driscoll this would be dope to have

david-driscoll:
This will have to come with #186  I'm out of work for most of the next week so I'll take a stab at finishing the 3.15 upgrade.

PR #227, opened 2020-04-09
Added two NuGet badges.

lextm:
It is not quite obvious what NuGet packages are produced from this repo, so added two badges for the commonly used ones.

Don't know


david-driscoll:
thanks @lextm !

PR #228, opened 2020-04-10
Rework ProcessScheduler

david-driscoll:
This is to rework the scheduler to achieve a few things:

1) The code has always been a pain point, it works but it's hard to work on.
2) We've never been able to put a hard cap on total number of parallel operations that can happen at a time (number of concurrent requests)

If you don't like observables cover your eyes.
If you like observables and know a slightly better way to fan-in fan-out let me know! 😄 

@TylerLeonhardt what are your thoughts on making completion resolve handler parallel?  Originally I had it serial because it can contain actions and edits but that would get invalidated by another document change anyway.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -33,6 +33,7 @@
     <PackageReference Update="coverlet.collector" Version="1.2.0" />
     <PackageReference Update="coverlet.msbuild" Version="2.8.0" />
     <PackageReference Update="System.Reactive" Version="4.3.2" />
+    <PackageReference Update="Microsoft.Reactive.Testing" Version="4.3.2" />
     <PackageReference Update="MediatR" Version="8.0.1" />
     <PackageReference Update="MediatR.Extensions.Microsoft.DependencyInjection" Version="8.0.0" />
   </ItemGroup>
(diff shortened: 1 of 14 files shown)


TylerLeonhardt:
well, if I hold down arrow, and go through my list of completions, that's a lot of stuff that would be happening serially.

In our case, it would be better parallel.. but you may want to ask @NTaylorMullen or others if that's ok.

NTaylorMullen:
Does this only impact scenarios where you return isInComplete: true?

/cc @ToddGrun @jimmylewis @alexgav

david-driscoll:
@NTaylorMullen so whenever the editor highlights an completion item, resolve is called to get additional details.  Partial lists are a different thing.

NTaylorMullen:
Ah what would the potential drawbacks be of it being parallel?

TylerLeonhardt:
> Ah what would the potential drawbacks be of it being parallel?

I think @david-driscoll's issue was that it has the ability to change state I think...:

```
/**
	 * An optional command that is executed *after* inserting this completion. *Note* that
	 * additional modifications to the current document should be described with the
	 * additionalTextEdits-property.
	 */
	command?: Command;
```
This is found on `CompletionItem`. However, the `Command` is executed after it's accepted... I don't really see any issues making it parallel.

NTaylorMullen:
> I think @david-driscoll's issue was that it has the ability to change state I think...:

I see. Ya I wouldn't think that making completion resolve parallel would cause any issues. That being said, in the middle of a completion session I wouldn't expect many other events to be going through the language server either though so i'm not sure it being parallel would buy a whole lot either

PR #229, opened 2020-04-10
Fix StackOverflowException in ProgressToken.Equals

instantsc:
Copy-pasting the server example causes the app to crash with a StackOverflowException. After hours of trying to debug this, i tracked the issue down to these lines of code. Would appreciate it if the problem was fixed one way or another


david-driscoll:
Thanks @instantsc the change makes sense.  Sorry for the headache debugging it!

PR #230, opened 2020-04-12
Added proposed features from lsp 3.16

david-driscoll:
This adds the current proposals for the next version of the language server protocol.

I'm introducing these with `[Obsolete]` attribute to indicate that they are preview and they may change in the future.  When 3.16 comes out we can simply remove the attributes and move the namespaces.

I got this working with a nice disco look in regards to semantic tokens 😄 
image


mholo65:
> I'm introducing these with [Obsolete] attribute to indicate that they are preview and they may change in the future. When 3.16 comes out we can simply remove the attributes and move the namespaces.

If there only was a `[Preview]` attribute with a analyzer that warns that these are preview 😀 Obsole feels wrong in this case.

david-driscoll:
@mholo65 maybe we should make one! Until then for now it'll at least point out that these bits may change or many not work without extra work.

src/Protocol/Client/Capabilities/DocumentSymbolCapability.cs
```diff
@@ -17,5 +17,16 @@ public class DocumentSymbolCapability : DynamicCapability, ConnectedCapability<I
         /// </summary>
         [Optional]
         public bool? HierarchicalDocumentSymbolSupport { get; set; }
+
+        /// <summary>
+        /// The client supports tags on `SymbolInformation`.Tags are supported on
+        /// `DocumentSymbol` if `hierarchicalDocumentSymbolSupport` is set tot true.
```
TylerLeonhardt, line 23:
```suggestion
        /// `DocumentSymbol` if `hierarchicalDocumentSymbolSupport` is set to true.
```

src/Protocol/Client/Capabilities/WorkspaceSymbolCapability.cs
```diff
@@ -10,5 +11,16 @@ public class WorkspaceSymbolCapability : DynamicCapability, ConnectedCapability<
         /// </summary>
         [Optional]
         public SymbolKindCapability SymbolKind { get; set; }
+
+        /// <summary>
+        /// The client supports tags on `SymbolInformation`.Tags are supported on
+        /// `DocumentSymbol` if `hierarchicalDocumentSymbolSupport` is set tot true.
```
TylerLeonhardt, line 17:
```suggestion
        /// `DocumentSymbol` if `hierarchicalDocumentSymbolSupport` is set to true.
```

src/Protocol/Document/Server/Proposals/ISemanticTokensEditsHandler.cs
```diff
@@ -0,0 +1,150 @@
...
+        public virtual async Task<SemanticTokensOrSemanticTokensEdits> Handle(SemanticTokensEditsParams request,
+            CancellationToken cancellationToken)
+        {
+            var document = await GetSemanticTokensDocument(request, cancellationToken);
+            var builder = document.Edit(request);
+            await Tokenize(builder, request, cancellationToken);
+            return builder.Commit().GetSemanticTokensEdits();
+        }
+
+        // TODO: Is this correct?
```
TylerLeonhardt, line 67: is it? :)

src/Protocol/DocumentUri.cs
```diff
@@ -0,0 +1,93 @@
+﻿using System;
+using System.IO;
+using OmniSharp.Extensions.LanguageServer.Protocol.Models;
+
+namespace OmniSharp.Extensions.LanguageServer.Protocol
+{
+    /// <summary>
+    ///     Helper methods for working with LSP document URIs.
+    /// </summary>
+    public static class DocumentUri
```
TylerLeonhardt, line 10: This is an interesting class. Would it handle VS Code's conflicting `Uri` type (aka the fact that `C:` is encoded to `C%3A`)

src/Protocol/Models/Position.cs
```diff
@@ -12,20 +12,20 @@ public Position()
...
         /// Line position in a document (zero-based).
         /// </summary>
-        public long Line { get; set; }
+        public int Line { get; set; }
 
         /// <summary>
         /// Character offset on a line in a document (zero-based).
         /// </summary>
-        public long Character { get; set; }
+        public int Character { get; set; }
```
TylerLeonhardt, line 28: any reason for changing all the `long`s to `int`s?

TylerLeonhardt:
I assume that this should "just work" with language servers that only support 3.15 handlers and also clients that only support 3.15-styled messages.

src/Protocol/Models/Position.cs
david-driscoll: To make it easier to deal with the math when generating semantic tokens, otherwise a lot of potential conversions can happen.  I can try and roll it back if it is needed for something.  I dunno how many files have 2GB characters on a single line.

src/Protocol/DocumentUri.cs
david-driscoll: We probably should add that here with some unit tests to make sure.

This was brought up from the client I think.

src/Protocol/Document/Server/Proposals/ISemanticTokensEditsHandler.cs
david-driscoll: I think that was from before I had added the helpers for tokenization.  At this point I think it's good based on my testing, and if it isn't I made it virtual for a reason 😄

david-driscoll:
Yeah, the goal here is existing servers will work, and if implement the new features and run into an editor that also implements the new features, things will "just work".

src/Protocol/Models/Position.cs
TylerLeonhardt: PowerShell uses ints so I'll be happy to remove a bunch of `(int)`s all over the place :)

src/JsonRpc/OutputHandler.cs
```diff
@@ -72,15 +75,18 @@ private void ProcessOutputQueue()
...
+                _outputHandler.LogTrace(ex, "Cancellation happened");
+                _outputIsFinished.TrySetException(ex);
+                // _thread.Abort(ex);
                 // else ignore. Exceptions: OperationCanceledException - The CancellationToken has been canceled.
             }
             catch (Exception e)
             {
+                _outputHandler.LogTrace(e, "Could not write to output handler, perhaps serialization failed?");
                 _outputIsFinished.TrySetException(e);
+                // _thread.Abort(e);
```
TylerLeonhardt, line 89: commented out code?

src/JsonRpc/JsonRpc.csproj
```diff
@@ -7,6 +7,7 @@
         <PackageDescription>Primitives for working with JsonRpc.  This library is used as the base for communication with language servers</PackageDescription>
     </PropertyGroup>
     <ItemGroup>
+        <PackageReference Include="System.Collections.Immutable" />
```
TylerLeonhardt, line 10: Curious what this is good for over `IReadOnly*` types.

src/Protocol/DocumentUri.cs
TylerLeonhardt: In those tests, can you also include tests of Chinese characters or Cyrillic or something? Those don't always play nicely with the `Uri` type...

src/JsonRpc/JsonRpc.csproj
david-driscoll: I grabbed it for ImmutableArray mainly because it "does the right thing" when it comes to creating new and transformed arrays.

I used this so I don't have to think too hard about copying arrays around.  This is because `SemanticTokensDocument` keeps a copy of the previous tokens, and the newly computed tokens, so that it can handle computing the diff between the two.

I was porting the builder from https://github.com/microsoft/vscode-languageserver-node/blob/master/server/src/semanticTokens.proposed.ts#L45 😄

PR #233, opened 2020-04-19
Add SendEmptyRequest Method for shutdown request

luxxia702:
The LanguageClient sends a Shutdown notification. But the LanguageServer expects a request (as per spec), causing a server-side deserialization error.


david-driscoll:
thanks @luxxia702

PR #235, opened 2020-05-03
Refactored IResponseRouter for better type inference

david-driscoll:
This is a quick change related `IResponseRouter`.

Currently if you have a params object and a response object you always have to define the type of both to call `SendRequest`.  This new pattern returns an object lets you define the return type of the request, so it's a little easier to do `SendRequest("somemethod", someParams).Returning<ReturnType>(cancellationToken)`

### DIFF
--- a/src/JsonRpc/IResponseRouter.cs
+++ b/src/JsonRpc/IResponseRouter.cs
@@ -9,12 +9,10 @@ namespace OmniSharp.Extensions.JsonRpc
     {
         void SendNotification(string method);
         void SendNotification<T>(string method, T @params);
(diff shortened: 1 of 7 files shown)


david-driscoll:
I've never been happy with the experience here.

```
.SendRequest<TRequest, TResponse>("method", params, token);
.SendRequest<TResponse>("method", token);
.SendRequest("method", params, token);
.SendRequest("method", token);
```

Those have all worked but the case where TRequest is required when you know what TRequest has always been painful, I always have to double back to the front of the method and add the correct type information.

PR #239, opened 2020-05-09
Added support for more generic uris

### LINKED ISSUE #238: Untitled files aren't handled correctly in the new DocumentUri class
VS Code likes to use `Untitled:Untitled-1` for the `Uri` of their untitled files... with the new `DocumentUri` class this is translated to:

```
> OmniSharp.Extensions.LanguageServer.Protocol.DocumentUri.Parse("Untitled:Untitled-1")
{file:///Untitled:Untitled-1}
Authority [string]:""
Fragment [string]:""
Path [string]:"/Untitled:Untitled-1"
Query [string]:""
Scheme [string]:"file"
Static members
```

This should be:

```
OmniSharp.Extensions.LanguageServer.Protocol.DocumentUri.Parse("Untitled:Untitled-1")
{file:///Untitled:Untitled-1}
Authority [string]:""
Fragment [string]:""
Path [string]:"Untitled-1"
Query [string]:""
Scheme [string]:"Untitled"
Static members
```

so that when you convert it to a string it goes to `Untitled:Untitled-1`

david-driscoll:
Fixes #238

### DIFF
--- a/src/Protocol/DocumentUri.cs
+++ b/src/Protocol/DocumentUri.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
+using System.Text.RegularExpressions;
 using OmniSharp.Extensions.LanguageServer.Protocol.Models;
 using static System.IO.Path;
 
@@ -14,17 +15,23 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol
     /// <remarks>This exists because of some non-standard serialization in vscode around uris and .NET's behavior when deserializing those uris</remarks>
     public class DocumentUri : IEquatable<DocumentUri>
     {
+        public static Regex WindowsPath =
+            new Regex("^\\w(?:\\:|%3a)[\\\\|\\/]", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
+        public string _delimiter = SchemeDelimiter;
+
         /// <summary>
         /// Create a new document uri
         /// </summary>
-        /// <param name="url"></param>
+        /// <param name="url"></param> add .
         public DocumentUri(string url)
         {
+            var uncMatch= false;
             var delimiterIndex = url.IndexOf(SchemeDelimiter, StringComparison.Ordinal);
-            if (delimiterIndex == -1)
+            if ((uncMatch = url.StartsWith("\\\\")) || (url.StartsWith("/")) || (WindowsPath.IsMatch(url)))
             {
                 // Unc path
-                if (url.StartsWith("\\\\"))
+                if (uncMatch)
                 {
                     var authorityEndIndex = url.IndexOf('\\', 2);
                     Authority = url.Substring(2, authorityEndIndex - 2);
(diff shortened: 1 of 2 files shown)


src/Protocol/DocumentUri.cs
```diff
@@ -14,17 +15,23 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol
...
+            new Regex("^\\w(?:\\:|%3a)[\\\\|\\/]", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
+        public string _delimiter = SchemeDelimiter;
+
         /// <summary>
         /// Create a new document uri
         /// </summary>
-        /// <param name="url"></param>
+        /// <param name="url"></param> add .
         public DocumentUri(string url)
```
mholo65, line 27: Remove

src/Protocol/DocumentUri.cs
```diff
@@ -14,17 +15,23 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol
     /// <remarks>This exists because of some non-standard serialization in vscode around uris and .NET's behavior when deserializing those uris</remarks>
     public class DocumentUri : IEquatable<DocumentUri>
     {
+        public static Regex WindowsPath =
+            new Regex("^\\w(?:\\:|%3a)[\\\\|\\/]", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
mholo65, line 19: Use verbatim strings (`@"..."`) for Regex expressions, it makes them more readable.

david-driscoll: that is a very good idea... wish I would have thought of that...

src/Protocol/DocumentUri.cs
david-driscoll: NO THE COMMENT STAGES THE CODE AUTOMATICALLY.... wait... nevermind! 😄

mholo65: Good for you it wasn't your password and just `git add .` that slipped :smile:

src/Protocol/DocumentUri.cs
```diff
@@ -14,17 +15,23 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol
     /// <remarks>This exists because of some non-standard serialization in vscode around uris and .NET's behavior when deserializing those uris</remarks>
     public class DocumentUri : IEquatable<DocumentUri>
     {
+        public static Regex WindowsPath =
```
TylerLeonhardt, line 18: Did you craft this or did you get it from somewhere? Surely something like this exists in the wild

src/Protocol/DocumentUri.cs
```diff
@@ -14,17 +15,23 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol
...
+            new Regex("^\\w(?:\\:|%3a)[\\\\|\\/]", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
+        public string _delimiter = SchemeDelimiter;
```
TylerLeonhardt, line 21: Should this be private? Because of the underscore

src/Protocol/DocumentUri.cs
```diff
@@ -14,17 +15,23 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol
...
         /// Create a new document uri
         /// </summary>
-        /// <param name="url"></param>
+        /// <param name="url"></param> add .
         public DocumentUri(string url)
         {
+            var uncMatch= false;
             var delimiterIndex = url.IndexOf(SchemeDelimiter, StringComparison.Ordinal);
-            if (delimiterIndex == -1)
+            if ((uncMatch = url.StartsWith("\\\\")) || (url.StartsWith("/")) || (WindowsPath.IsMatch(url)))
```
TylerLeonhardt, line 31: I'm trying to think if it's safer to only do this on Windows... Since UNC style paths are not a thing on unix.

src/Protocol/DocumentUri.cs
david-driscoll: probably... it was late at night

src/Protocol/DocumentUri.cs
david-driscoll: It doesn't matter because later on we special case unc later on in the flow.  Basically that check is to handle the well known non-uri paths that we can convert into a file uri.

PR #242, opened 2020-05-18
URL encode when running ToString()

TylerLeonhardt:
This `EscapeUriString` on the Path and
`EscapeDataString` on the Query and Fragment

It also updates all the tests to reflect that we will receive encoded urls.

### DIFF
--- a/test/JsonRpc.Tests/InputHandlerTests.cs
+++ b/test/JsonRpc.Tests/InputHandlerTests.cs
@@ -128,8 +128,10 @@ namespace JsonRpc.Tests
         }
 
         [Theory]
-        [InlineData("{\"changes\": [{\"uri\": \"file:///Mörkö.cs\",\"type\": 1}]}")]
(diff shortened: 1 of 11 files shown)


test/JsonRpc.Tests/InputHandlerTests.cs
```diff
@@ -128,8 +128,10 @@ public void ShouldPassInUtf8EncodedRequests()
         }
 
         [Theory]
-        [InlineData("{\"changes\": [{\"uri\": \"file:///Mörkö.cs\",\"type\": 1}]}")]
-        [InlineData("{\"textDocument\": {\"uri\": \"file://abc/123/树.cs\"}}")]
+        // Mörkö
```
filipw, line 131: I used to be really afraid of Mörkö as a kid

mholo65: Who hasn't been afraid of Mörkö? 😅

david-driscoll: rofl

PR #245, opened 2020-05-21
DocumentUri Version 2!

david-driscoll:
This should be mostly net new functionality from an api perspective.


TylerLeonhardt:
Here are some changes that _seem_ to pass all the tests on macOS:
https://github.com/OmniSharp/csharp-language-server-protocol/pull/246

PR #248, opened 2020-05-22
Overhaul part 2/2 (Refactored Language Client, Pipelines and unit tests!)

david-driscoll:
So yeah... this is huge, sorry for that.

This includes the pipeline changes (and revealed a few  problems with the `InputHandler`)

This rewrites the `LanguageClient` to be the `LanguageServer` just in reverse.  Same design but mirrored methods.  (ie. The server will `PublishDiagostics` and the client will have an `IPublishDiagnosticsHandler` to recieve them).

As a core part of this change the goal is to ensure consistency between handlers.  This has introduced a few things.

JsonRpc
* `MethodAttribute` has a new (optional) enum added to it for the `Direction` this handler, which is just used a hint for a JsonRpcServer.  Direction is required for both LSP and DAP protocols however, if it's missing there is a unit test which will fail.

LSP/DAP
* All request classes now require a method attribute.  This method must match the method defined on the handler.
* All method definitions must have a direction on them.
  * This used to ensure that there are appropriate methods and classes for both sending and recieving events.
* Extension Methods Refactored
  * Notifications are "fire an forget" for the caller, on the callee side they might take a while, so notifications have overloads for async and sync.
  
LSP
* Extension Methods Refactored
  * Requests that support `IPartialResultRequest` or `IPartialResultsRequest` have an overload for dealing with partial results
* Progress is now supported via the all new `IProgressManager`
  * The server or client and monitor results sent by the other side.
* Partial Results Progress is now supported by the new `IProgressManager` ( https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#partialResults )
  * The server implement logic using partial results even if the client does not support them, they will automatically be translated into the correct response object for you using `MonitorUntil`
  * The client will return a `TaskLike` object that is both an `IObservable<T>` and `{ GetAwaiter(): TaskAwaiter<TResponse> }` 
    * awaiting or observing should be transparent to the consumer.
* Work Done Progress is now supported bt the new `IServerWorkDoneManager` and `IClientWorkDoneManager` ( https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workDoneProgress )
  * There is an extension method that works with `IClientProxy` for sending requests, which will enrich the request with a work done token.
  * There is then a method on `IServerWorkDoneManager` that is safe to call on `IWorkDoneProgressParams` even if workdone is not set, it will just be an Observer that goes no where

LSP/DAP Unit Testing
* Unit testing is now a first class citizen with 3 new projects
  * OmniSharp.Extensions.JsonRpc.Testing - Enables base testing support for both a generic server (lsp/dap) and `JsonRpcServer`
  * OmniSharp.Extensions.DebugAdapter.Testing
  * OmniSharp.Extensions.LanguageProtocol.Testing
* Unit Test classes are framework agnostic (no dependency on XUnit by default)
  * Includes settings for test timeout (with a CancellationToken)
  * Includes the ability to wait for servers and/or clients to `Settle` to ensure state is consistent on both sides
  * Logging is supported on both sides, so it can be tied into framework specific features (ie via `ITestOutputHandler`)
  
With the new client we now have the ability to test the library in both directions to help improve code coverage as well as help with developing tests for corner cases (such as servers that don't support dynamic registration for example).


Directory.Build.targets
```diff
@@ -33,12 +34,15 @@
...
         <PackageReference Update="Microsoft.Reactive.Testing" Version="4.3.2"/>
         <PackageReference Update="MediatR" Version="8.0.1"/>
         <PackageReference Update="MediatR.Extensions.Microsoft.DependencyInjection" Version="8.0.0"/>
         <PackageReference Update="Bogus" Version="29.0.1"/>
         <PackageReference Update="Snapper" Version="2.2.4"/>
         <PackageReference Update="Xunit.SkippableFact" Version="1.4.8" />
-        <PackageReference Update="DynamicData" Version="6.14.18" />
+        <PackageReference Update="System.IO.Pipelines" Version="4.7.1"/>
+        <PackageReference Update="Nerdbank.Streams" Version="2.4.60" />
+        <!-- <PackageReference Update="DynamicData" Version="6.14.18" /> -->
```
TylerLeonhardt, line 46: Don't need this anymore?

PR #250, opened 2020-06-02
Add support for strongly typed data to be attached to */resolve

david-driscoll:
This is an initial stab at adding support for strongly typed data to be attached to codelens, completion items, and document links.   

Adds support for multiple codelens, completion and document link handlers for each text document.

What this means is that you can have multiple codelens (and completion/document link) handlers for a document selector.

You might have a codelens handler that returns test lenses (and sets up data specific to text lenses).  You could then have another handler that returns code lens for references, with a different data element.  These will then be automatically merged for you by the server, and resolve requests will be routed to the correct handler.


david-driscoll:
#269 superseeds this one, as it has fewer breaking changes.

PR #251, opened 2020-06-02
fixed an bug in AddTextDocumentIdentifier where it would not register…

david-driscoll:
… the correct type.


david-driscoll:
This an oops... 😢  will require a 0.17.1 sadly.

PR #256, opened 2020-06-10
Added code generation for OnXyz and SendXyz methods (1/2)

david-driscoll:
This is part one that only includes a few file changes to make the objective a little clearer.  the next is just a bunch of deleted code (yay deleted code! 🥳 )


david-driscoll:
TLDR: I'm sick of trying to keep all the methods in sync, but want to offer both an interface and delegate based interface for every handler.  This solves that since handlers follow a fairly common pattern.

NTaylorMullen:
Really cool idea! First seeing source generators in action. How was the dev experience with em?

david-driscoll:
@NTaylorMullen I was actually pretty pleased with it.  I have not played with the latest source generators but I assume they would be fairly similar.  Sometimes I'd break things literally in half, break the generator, and using generated methods and things fall apart fast... however that's what unit tests are for.

PR #263, opened 2020-07-22
Updated latest version of the 0.16 spec

david-driscoll:


src/Protocol/Models/InitializeParams.cs
```diff
@@ -57,7 +57,7 @@ public string RootPath
         /// The initial trace setting. If omitted trace is disabled ('off').
         /// </summary>
         [Optional]
-        public InitializeTrace Trace { get; set; } = InitializeTrace.Off;
+        public TraceValue TraceValue { get; set; } = TraceValue.Off;
```
mholo65, line 60: Will this rename break existing clients? Rename of property `Trace` to `TraceValue`.

david-driscoll: yes... that shouldn't be there, nice catch.  I renamed the enum to line up with the spec, but the property name should stay.

sample/SampleServer/Program.cs
```diff
@@ -40,15 +44,18 @@ static async Task MainAsync(string[] args)
...
                         .SetMinimumLevel(LogLevel.Debug))
                     .WithHandler<TextDocumentHandler>()
                     .WithHandler<DidChangeWatchedFilesHandler>()
                     .WithHandler<FoldingRangeHandler>()
                     .WithHandler<MyWorkspaceSymbolsHandler>()
                     .WithHandler<MyDocumentSymbolHandler>()
-                    .WithHandler<SemanticTokens>()
+                    .WithHandler<SemanticTokensHandler>()
+                    // Linking here because the spec changed but vscode is behind
+                    .WithLink(TextDocumentNames.SemanticTokensFull, "textDocument/semanticTokens")
```
david-driscoll, line 57: @TylerLeonhardt this lets us fake out the vscode client for now, until it's updated to line up with the spec.

TylerLeonhardt:
yay!

PR #267, opened 2020-07-28
Debug Adapter Server / Client update

david-driscoll:
This updates to the latest version of the Debug Adapter Protocol bringing in some of the missing pieces, cancellation, progress and a few other things.

Also adds support for a Server / Client pair exactly like Language Server Protocol has, these are super awesome in testing scenarios.  

* The initialization process is handled.
* Progress can be sent from the server to the client.
* Cancellation is now supported
  * This required some gymnastics to get right
  * The cancel handler (implemented by `IServerProgressManager` will get essentially empty requests for request cancellation, as they get re-routed down the normal cancellation path as provided by the rest of the json rpc bits.
* The data sent from the server to the client includes the LSP error codes, which are used to re-hydrate the proper exception types on the client side.

Minor change to some types (though they are really only used on the the internals) to include a method, this was to properly fix serialization of DAP responses.


david-driscoll:
cc @TylerLeonhardt @rjmholt

TylerLeonhardt:
Some things I wanna make sure are addressed are:

* How do we extend the launch and attach messages? Currently, we redefine our own type.https://github.com/PowerShell/PowerShellEditorServices/blob/master/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/LaunchAndAttachHandler.cs#L30-L95
* How could I disconnect and then later connect to a debug adapter server? Currently, we do this horrible event stuff where we:
    * Trigger [this event](https://github.com/PowerShell/PowerShellEditorServices/blob/master/src/PowerShellEditorServices/Server/PsesDebugServer.cs#L151-L154) [In the "disconnect" message](https://github.com/PowerShell/PowerShellEditorServices/blob/master/src/PowerShellEditorServices/Services/DebugAdapter/Handlers/DisconnectHandler.cs#L80)
    * [That Disposes of the old server and restarts it](https://github.com/PowerShell/PowerShellEditorServices/blob/f4c3f792029da222f70d4538350b2eba399f55c9/src/PowerShellEditorServices.Hosting/Internal/EditorServicesRunner.cs#L257-L267) (YUCK)

david-driscoll:
I knew there were pieces I need to make sure got updated, thanks I'll look at this tonight.

david-driscoll:
@TylerLeonhardt have any thoughts on how to make it so those requests are more flexible?

I can make it a Dictionary or add a `[JsonExtensionData]` property... however none of those will be "strongly" typed.  Re-defining it honestly seems like the better approach.  Perhaps if you derived from the requests themselves and added the properties you want....  I'm not sure.

I didn't get to the second point today, I'll look at that tomorrow.

TylerLeonhardt:
👀👀👀👀👀👀

PR #318, opened 2020-08-15
Remove jank

davidwengier:
Using GeneratorDriver means you don't need all of that jank.

This is slightly different because it will create a compilation, which the previous code didn't. I've left commented out the line that validates that resulting compilation because three of the tests fail with it left in, but I don't know enough about what the generator is supposed to do to know if thats valid or not.

Feel free to throw this away if it's not helpful :)


david-driscoll:
Thanks @davidwengier this was something I was going to get to!

I saw the twitter post yesterday and did this the night before I saw that. ❤️

PR #342, opened 2020-08-29
drop down to v2 of the Microsoft.Extenisons.*

david-driscoll:


NTaylorMullen:
@ryanbrandenburg once this is available for consumption we should change our O# version to this so we don't need to binding redirect the extension dependencies.

PR #344, opened 2020-09-01
Fixed DocumentTextSync converter / Added semantic token refresh notification

david-driscoll:


src/Protocol/LanguageProtocolDelegatingHandlers.cs
```diff
@@ -334,48 +334,48 @@ public RequestCapability(Guid id, Func<TParams, TCapability, CancellationToken,
             void ICapability<TCapability>.SetCapability(TCapability capability) => _capability = capability;
         }
 
-        public sealed class PartialResult<TItem, TResponse, TCapability, TRegistrationOptions> :
-            IJsonRpcRequestHandler<TItem, TResponse>,
+        public sealed class PartialResult<TParams, TResponse, TItem, TCapability, TRegistrationOptions> :
```
david-driscoll, line 337: These were wrong.  Since nothing had used them yet, we didn't know.

PR #354, opened 2020-09-09
Use ConfigureAwait when booting up input stream.

NTaylorMullen:
- Razor initializes O#'s language server in-proc in Visual Studio and because the `pipeReader.ReadAsync` doesn't utilize `ConfigureAwait(false)` it attempts to only ever read information on the main thread which is a problem if a feature such as light bulbs or go-to-definition require.


TanayParikh (approved):

david-driscoll (approved):
There are probably more places we could add this, but that can be handled later.

CI is not you, it's me, should be fixed shortly.

NTaylorMullen:
> There are probably more places we could add this, but that can be handled later.

ya I figured. I try to default to using ConfigureAwait(false) but to be frank this is the first time when it's actually ever bitten me 😄

david-driscoll:
So the test failures seem legit... why they're legit I do not know, however I will figure it out tonight.

PR #392, opened 2020-10-13
Custom handlers not override built in handlers correctly

david-driscoll:
* Fixed a bug in configuration (configuration values were case-sensitive).
* Added unit tests demonstrating Configuration.Binder, and Options usages

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -17,6 +17,9 @@
     <PackageReference Update="Microsoft.Extensions.Logging" Version="2.0.0" />
     <PackageReference Update="Microsoft.Extensions.Logging.Debug" Version="2.0.0" />
     <PackageReference Update="Microsoft.Extensions.Configuration" Version="2.0.0" />
+    <PackageReference Update="Microsoft.Extensions.Configuration.Binder" Version="2.0.0" />
+    <PackageReference Update="Microsoft.Extensions.Options" Version="2.0.0" />
+    <PackageReference Update="Microsoft.Extensions.Options.ConfigurationExtensions" Version="2.0.0" />
     <PackageReference Update="Microsoft.Extensions.DependencyInjection" Version="2.0.0" />
     <PackageReference Update="Newtonsoft.Json" Version="11.0.2" />
     <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.7.1" />
--- a/src/Client/LanguageClientOptionsExtensions.cs
+++ b/src/Client/LanguageClientOptionsExtensions.cs
@@ -108,18 +108,6 @@ namespace OmniSharp.Extensions.LanguageServer.Client
             return options;
         }
 
-        public static LanguageClientOptions UseDefaultRegistrationManager(this LanguageClientOptions options, bool value = true)
-        {
-            options.DefaultRegistrationManager = value;
-            return options;
-        }
-
-        public static LanguageClientOptions UseDefaultWorkspaceFolderManager(this LanguageClientOptions options, bool value = true)
-        {
-            options.DefaultWorkspaceFolderManager = value;
-            return options;
-        }
-
         public static LanguageClientOptions ConfigureLogging(
             this LanguageClientOptions options,
             Action<ILoggingBuilder> builderAction
--- a/src/Client/LanguageClientOptionsExtensions.cs
+++ b/src/Client/LanguageClientOptionsExtensions.cs
@@ -108,6 +108,18 @@ namespace OmniSharp.Extensions.LanguageServer.Client
             return options;
         }
 
+        public static LanguageClientOptions UseDefaultRegistrationManager(this LanguageClientOptions options, bool value = true)
+        {
+            options.DefaultRegistrationManager = value;
+            return options;
+        }
+
+        public static LanguageClientOptions UseDefaultWorkspaceFolderManager(this LanguageClientOptions options, bool value = true)
+        {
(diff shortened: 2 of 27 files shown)


src/Client/LanguageClientOptionsExtensions.cs
```diff
@@ -108,6 +108,18 @@ public static LanguageClientOptions OnStarted(this LanguageClientOptions options
             return options;
         }
 
+        public static LanguageClientOptions UseDefaultRegistrationManager(this LanguageClientOptions options, bool value = true)
```
NTaylorMullen, line 111: Hmm, feels interesting to have `UseDefaultXYZ` isn't that inferred at the services collection layer by whether or not a consumer registered their own types? Ditto question for the workspace folder manager

Directory.Build.targets
```diff
@@ -17,6 +17,9 @@
     <PackageReference Update="Microsoft.Extensions.Logging" Version="2.0.0" />
     <PackageReference Update="Microsoft.Extensions.Logging.Debug" Version="2.0.0" />
     <PackageReference Update="Microsoft.Extensions.Configuration" Version="2.0.0" />
+    <PackageReference Update="Microsoft.Extensions.Configuration.Binder" Version="2.0.0" />
```
NTaylorMullen, line 20: What do the new dependencies enable from a consumers perspective?

src/Client/LanguageClientOptionsExtensions.cs
david-driscoll: They may actually not be needed, I'm working through the underlying issue.

Directory.Build.targets
david-driscoll: lights up some of the options behavior, and the new configure methods, but strictly isn't needed.  And I'm not sure I want to bring these changes in with a patch.

david-driscoll: So these are in here, however the example only exists in the unit tests.  It would be on the consumer to pull them in until the next version.

src/JsonRpc/JsonRpcServerServiceCollectionExtensions.cs
```diff
@@ -79,23 +81,51 @@ internal static IContainer AddJsonRpcMediatR(this IContainer container)
...
+            public Task<R> Handle(T request, CancellationToken cancellationToken)
+            {
+                if (_requestContext == null)
+                {
+                    if (_handler == null)
+                    {
+                        throw new NotImplementedException($"No request handler was registered for type {typeof(IRequestHandler<T, R>).FullName}");
 
-            return container.With(
-                rules => rules.WithUnknownServiceResolvers(
```
david-driscoll, line 84: The fundamental problem was, the "built-in" handlers are getting added to the container, so they are never "unknown", and therefore this code doesn't run.

src/JsonRpc/JsonRpcServerServiceCollectionExtensions.cs
```diff
@@ -79,23 +81,51 @@ internal static IContainer AddJsonRpcMediatR(this IContainer container)
             container.RegisterMany(new[] { typeof(IMediator).GetAssembly() }, Registrator.Interfaces, Reuse.ScopedOrSingleton);
             container.RegisterMany<RequestContext>(Reuse.Scoped);
             container.RegisterDelegate<ServiceFactory>(context => context.Resolve, Reuse.ScopedOrSingleton);
+            container.Register(typeof(IRequestHandler<,>), typeof(RequestHandler<,>));
```
david-driscoll, line 84: This replacement code ensure that handlers can be added to the container as their MediatR interface `IRequestHandler<,>` and that will override the default registration, this is fine.
The decorator will ensure that even if a handler is registered that the handler from the current request will be used instead.

Directory.Build.targets
NTaylorMullen: Without them are consumers of the O# framework library still able to bring them in themselves and reap the benefits if they write the code themselves? Wondering if it adds a lot of value to have them part of the base framework is all

david-driscoll: I think it has value, however I want to minimize the impact so this can be released as a patch fairly soon.

ajaybhargavb (commented):
I just tried this out. The didChangeConfiguration requests are coming through to the server now yay! However, I noticed that the response received by the server for `workspace/configuration` requests aren't correct. No matter what setting I set on the client, the result from this request is always the default value for each property. Calling `vscode.workspace.getConfiguration()` on the client returns the correct values so this isn't an issue with the client.

Another thing I noticed after upgrading to your latest commit is that I keep seeing a bunch of these exceptions. They weren't happening before afaik,
image

global.json
```diff
@@ -1,7 +0,0 @@
-{
```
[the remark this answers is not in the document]
david-driscoll: There was a build issue, I mainly had it in there for local development (since I have the .net 5 installed) but it isn't strictly needed for CI.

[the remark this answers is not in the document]
ajaybhargavb: I see. I asked because when I tried to build locally with a .net 5 sdk, it failed. I had to pin it to 3.1.402 to make it work.

david-driscoll:
I worked with @ajaybhargavb and I think we solved his underlying problem, and it looks like there are no more changes required.  however I'll wait until I hear back if any other problems popped up.

ajaybhargavb (approved):
LGTM. I tested it out and looks like this fixed the issue. Thanks @david-driscoll!

PR #401, opened 2020-10-28
react to new registration type for semantic tokens

### LINKED ISSUE #397: Should the registerCapability request refer to the semantic tokens capability as `textDocument/semanticTokens` or `textDocument/semanticTokens/full`?
We're using Omnisharp.Extensions.LanguageServer (which is great, by the way!) to build a language server which is consumed by the vscode-languageclient NPM package. Recently we were upgraded from [7.0.0-next.10](https://www.npmjs.com/package/vscode-languageclient/v/7.0.0-next.10) to [7.0.0-next.12](https://www.npmjs.com/package/vscode-languageclient/v/7.0.0-next.12), which has renamed the `textDocument/semanticTokens/full` capability to `textDocument/semanticTokens`, and hits an error when the Omnisharp server calls `client/registerCapability` with the former. It also seems like the LSP spec has been recently updated accordingly.

Do you happen to know what the correct behavior is? If it looks more like an issue with the client library, feel free to close this!

david-driscoll:
Fixes #397 

Reference commit for latest vscode client: https://github.com/microsoft/vscode-languageserver-node/commit/bb85e80617c7f0cb64dad9dad852219e2da7c4c9
LSP reference: https://github.com/microsoft/language-server-protocol/issues/1107

### DIFF
--- a/src/Protocol/Document/Proposals/ISemanticTokensHandler.cs
+++ b/src/Protocol/Document/Proposals/ISemanticTokensHandler.cs
@@ -27,7 +27,7 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol.Document.Proposals
     [GenerateHandlerMethods]
     public interface ISemanticTokensDeltaHandler :
         IJsonRpcRequestHandler<SemanticTokensDeltaParams, SemanticTokensFullOrDelta?>,
-        IRegistration<SemanticTokensRegistrationOptions>, ICapability<SemanticTokensCapability>, IDoesNotParticipateInRegistration
+        IRegistration<SemanticTokensRegistrationOptions>, ICapability<SemanticTokensCapability>
     {
     }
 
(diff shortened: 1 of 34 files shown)


src/Protocol/Document/Proposals/ISemanticTokensHandler.cs
```diff
@@ -36,7 +36,7 @@ public interface ISemanticTokensDeltaHandler :
     [Method(TextDocumentNames.SemanticTokensRange, Direction.ClientToServer)]
     [GenerateHandlerMethods]
     public interface ISemanticTokensRangeHandler : IJsonRpcRequestHandler<SemanticTokensRangeParams, SemanticTokens?>,
-                                                   IRegistration<SemanticTokensRegistrationOptions>, ICapability<SemanticTokensCapability>, IDoesNotParticipateInRegistration
```
david-driscoll, line 39: These are just marker interfaces, so this shouldn't be a breaking change.

david-driscoll:
cc @anthony-c-martin

david-driscoll:
cc @NTaylorMullen @ajaybhargavb

david-driscoll:
Merging in for now, no real breaking changes except for the new client that causes issues.

PR #415, opened 2020-11-07
Updated to the latest spec changes

david-driscoll:


src/Protocol/Models/InsertReplaceEdit.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace OmniSharp.Extensions.LanguageServer.Protocol.Models
+{
+    /// <summary>
+    /// A special text edit to provide an insert and a replace operation.
+    ///
+    /// @since 3.16.0 - proposed state
+    /// </summary>
+    [DebuggerDisplay("{" + nameof(DebuggerDisplay) + ",nq}")]
+    public class InsertReplaceEdit
```
david-driscoll, line 11: This will be added onto the `CompletionItem` but this is a breaking change so this has not been put into place yet.

PR #416, opened 2020-11-07
FEEDBACK: Experimenting with records

david-driscoll:
I'm looking for feedback here for a future version.

Goals
* Use records where ever possible
* Make all the models Immutable with `init` setters.
  * models can be changed with the `with` operator.

Requirements
* Still support .NET Standard 2.0
  * This looks to be working
* Will require C# 9
  * Requires the .NET 5 SDK to build
  * Still will support .NET Standard 2.0

### DIFF
--- a/src/JsonRpc/Nullable/RecordAttributes.cs
+++ b/src/JsonRpc/Nullable/RecordAttributes.cs
@@ -1,0 +1,23 @@
+#pragma warning disable MA0048 // File name must match type name
+#define INTERNAL_RECORD_ATTRIBUTES
+#if NETSTANDARD2_0 ||  NETSTANDARD2_1 ||  NETCOREAPP2_0 ||  NETCOREAPP2_1 ||  NETCOREAPP2_2 || NET45 || NET451 || NET452 || NET6 || NET461 || NET462 || NET47 || NET471 || NET472 || NET48
+using System.ComponentModel;
+
+// ReSharper disable once CheckNamespace
+namespace System.Runtime.CompilerServices
+{
+    /// <summary>
+    /// Reserved to be used by the compiler for tracking metadata.
+    /// This class should not be used by developers in source code.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+#if INTERNAL_RECORD_ATTRIBUTES
+    internal
+#else
+    public
+#endif
+    static class IsExternalInit
+    {
+    }
+}
+#endif
--- a/src/Protocol/Document/TextDocumentAttributes.cs
+++ b/src/Protocol/Document/TextDocumentAttributes.cs
@@ -2,8 +2,10 @@ using System;
 
 namespace OmniSharp.Extensions.LanguageServer.Protocol.Document
 {
-    public class TextDocumentAttributes : IEquatable<TextDocumentAttributes>
+    public record TextDocumentAttributes
     {
+        public TextDocumentAttributes() { }
+
         public TextDocumentAttributes(DocumentUri uri, string languageId)
         {
             Uri = uri;
@@ -17,38 +19,8 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol.Document
             LanguageId = languageId;
         }
 
-        public DocumentUri Uri { get; }
-        public string? Scheme { get; }
-        public string LanguageId { get; }
-
-        public bool Equals(TextDocumentAttributes? other)
-        {
-            if (ReferenceEquals(null, other)) return false;
-            if (ReferenceEquals(this, other)) return true;
-            return Uri.Equals(other.Uri) && Scheme == other.Scheme && LanguageId == other.LanguageId;
-        }
-
-        public override bool Equals(object? obj)
-        {
-            if (ReferenceEquals(null, obj)) return false;
-            if (ReferenceEquals(this, obj)) return true;
-            if (obj.GetType() != GetType()) return false;
-            return Equals((TextDocumentAttributes) obj);
-        }
-
-        public override int GetHashCode()
-        {
-            unchecked
-            {
-                var hashCode = Uri.GetHashCode();
-                hashCode = ( hashCode * 397 ) ^ ( Scheme != null ? Scheme.GetHashCode() : 0 );
-                hashCode = ( hashCode * 397 ) ^ LanguageId.GetHashCode();
-                return hashCode;
-            }
-        }
-
-        public static bool operator ==(TextDocumentAttributes? left, TextDocumentAttributes? right) => Equals(left, right);
-
-        public static bool operator !=(TextDocumentAttributes? left, TextDocumentAttributes? right) => !Equals(left, right);
+        public DocumentUri Uri { get;init; }
+        public string? Scheme { get; init; }
+        public string LanguageId { get;init; }
     }
 }
--- a/src/Protocol/DocumentUri.cs
+++ b/src/Protocol/DocumentUri.cs
@@ -28,7 +28,7 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol
     /// </summary>
     /// <remarks>This exists because of some non-standard serialization in vscode around uris and .NET's behavior when deserializing those uris</remarks>
     [JsonConverter(typeof(DocumentUriConverter))]
-    public partial class DocumentUri : IEquatable<DocumentUri?>
+    public partial record DocumentUri
     {
         /// <summary>
         /// scheme is the "http' part of 'http://www.msft.com/some/path?query#fragment".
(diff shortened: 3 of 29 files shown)


david-driscoll:
cc @NTaylorMullen @TylerLeonhardt @ryanbrandenburg (and others)

src/Protocol/Document/TextDocumentAttributes.cs
```diff
@@ -2,8 +2,10 @@
 
 namespace OmniSharp.Extensions.LanguageServer.Protocol.Document
 {
-    public class TextDocumentAttributes : IEquatable<TextDocumentAttributes>
+    public record TextDocumentAttributes
     {
+        public TextDocumentAttributes() { }
```
ryanbrandenburg, line 7: Is there a particular reason we need the empty constructor?

src/Protocol/Document/TextDocumentAttributes.cs
```diff
@@ -2,8 +2,10 @@
...
+
         public TextDocumentAttributes(DocumentUri uri, string languageId)
```
ryanbrandenburg, line 9: This doesn't [seem to be the idiomatic way](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#record-types) of assigning properties in a record, though it will probably work as is, so it's mostly a matter of preference (I'm personally loving the typescript-y property assignment/definition.

src/Protocol/Models/PlaceholderRange.cs
```diff
@@ -1,8 +1,7 @@
 namespace OmniSharp.Extensions.LanguageServer.Protocol.Models
 {
-    public class PlaceholderRange
+    public record PlaceholderRange(Range Range, string Placeholder)
```
ryanbrandenburg, line 3: I love this new property definition style.

src/Protocol/Models/TextDocumentIdentifier.cs
```diff
@@ -4,44 +4,13 @@
...
-        public DocumentUri Uri { get; set; } = null!;
-
-        public bool Equals(TextDocumentIdentifier? other)
-        {
-            if (ReferenceEquals(null, other)) return false;
-            if (ReferenceEquals(this, other)) return true;
-            return Uri.Equals(other.Uri);
-        }
-
-        public override bool Equals(object? obj)
```
ryanbrandenburg, line 27: Lots of boilerplate being removed here, pretty nice considering I believe that was a big goal of records.

src/JsonRpc/Nullable/RecordAttributes.cs
```diff
@@ -0,0 +1,23 @@
...
+
+// ReSharper disable once CheckNamespace
+namespace System.Runtime.CompilerServices
+{
+    /// <summary>
+    /// Reserved to be used by the compiler for tracking metadata.
+    /// This class should not be used by developers in source code.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+#if INTERNAL_RECORD_ATTRIBUTES
```
NTaylorMullen, line 14: lol What's this about?

src/Protocol/Models/VersionedTextDocumentIdentifier.cs
```diff
@@ -4,43 +4,15 @@
 namespace OmniSharp.Extensions.LanguageServer.Protocol.Models
 {
     [DebuggerDisplay("{" + nameof(DebuggerDisplay) + ",nq}")]
-    public class VersionedTextDocumentIdentifier : TextDocumentIdentifier, IEquatable<VersionedTextDocumentIdentifier>
+    public record VersionedTextDocumentIdentifier(DocumentUri Uri, int Version = 0) : TextDocumentIdentifier(Uri)
```
ryanbrandenburg, line 7: There seem to be some behavior changes slipping in here (Version is now int instead of int?, defaults to 0 instead of null). That's chill if they were considered and intended, but personally I try to separate taking advantage of new features and making code-surface/behavior changes to avoid unexpected side-effects.

test/Lsp.Tests/SemanticTokensDocumentTests.cs
```diff
@@ -126,31 +126,13 @@ public class ReturnDocumentTokensFromScratchForRangeData : TheoryData<Range, IEn
...
-                        Start = new Position {
-                            Line = 12,
-                            Character = 21,
-                        },
-                        End = new Position {
-                            Line = 14,
-                            Character = 27
-                        }
-                    },
+                    ((12, 21),(14, 27)),
```
ryanbrandenburg, line 129: Hmm, C#9 certainly makes this more terse, but I'd have to hover over `Add` to find out that this is `Range` and `Position`. Personally I'd make that more explicit, but it's not my code-base.

src/Protocol/DocumentUri.cs
```diff
@@ -28,7 +28,7 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol
     /// </summary>
     /// <remarks>This exists because of some non-standard serialization in vscode around uris and .NET's behavior when deserializing those uris</remarks>
     [JsonConverter(typeof(DocumentUriConverter))]
-    public partial class DocumentUri : IEquatable<DocumentUri?>
+    public partial record DocumentUri
```
NTaylorMullen, line 31: Man it's pretty dope that records just make things equatable

ryanbrandenburg (approved):
Couple of style opinions and pointing out some things you probably already know about type changes. Nothing to block on if you've already considered these things.

Now I kinda want to do a similar PR in aspnetcore-tooling!

NTaylorMullen (approved):
Oh man, this PR definitely made me appreciate records more. Started playing around with them more 😄

src/JsonRpc/Nullable/RecordAttributes.cs
david-driscoll: this is an attribute the compiler needs (that exists on in `net5.0`) to build records correctly, while still staying .NET Standard 2.0 / 2.1 Compatible.  It's similar to how you can include types to get nullable reference types to work correctly with .NET Standard 2.0

src/Protocol/Models/VersionedTextDocumentIdentifier.cs
david-driscoll: Yeah this will be a breaking change, honestly it probably never should have been nullable, but that's how it is for the moment.

david-driscoll:
To be clear this isn't the final set of changes, I just want some feedback as to if the idea makes sense, how much pain it might cause, etc.-

src/Protocol/Document/TextDocumentAttributes.cs
david-driscoll: Backwards compatibility

src/Protocol/Document/TextDocumentAttributes.cs
david-driscoll: There are some JSON deserialization issues which is why some of these types have default constructors as well, so we might not get all the wins of a records, at least until we can move to System.Text.Json.  SJT is going to be a ways away for now.

src/Protocol/DocumentUri.cs
david-driscoll: and actually... this might need to revert this..  because of case sensitive file systems.

test/Lsp.Tests/SemanticTokensDocumentTests.cs
david-driscoll: that's actually the implicit operators, but there is nothing stopping you from using which syntax makes sense.

src/Protocol/Document/TextDocumentAttributes.cs
david-driscoll: oh and json serialization I found out later, lol.

ryanbrandenburg: Ah, yeah I've run into the json thing before, it's a real pain.

david-driscoll:
Any issues if anything that goes records has `init` only properties?

NTaylorMullen:
> Any issues if anything that goes records has `init` only properties?

Don't think so but I'm unfamiliar with the runtime implications of `init`.

david-driscoll:
I don't think there are any runtime implications, I think it's all compiler.  From what I've seen in reflection and generated code they appear to generate out the same, and act very similarly to Getter only properties, the only difference is the compiler lets you set them during object initialization but not after.

NTaylorMullen:
> I don't think there are any runtime implications

Ya I wouldn't think so either. If that's the case we should be good to go though! Records 4 life

PR #432, opened 2020-11-18
Breaking: LSP ISerializer interface is now deprecated, and marked obsolete

david-driscoll:
* LSP Serializer renamed to LSPSerializer
* Added serialization rules for client
  * Ensure code actions are provided properly.
* Updated serialization rules for server


david-driscoll:
Generally speaking this shouldn't break anyone, the custom serializer is used to manage the supported client/server tags, symbols, etc.

PR #440, opened 2020-11-25
Breaking: Registration Options require Client Capability during creation

david-driscoll:
## Registration Changes
The 3.16 specification defines the SemanticLegend in such a way that the client produces a list of supported tokens and the server must return the tokens it supports and the index order it will use.

Today we get the registration options at registration of a handler, but this is typically before we have started communicating with the server.

This change adds a new interface `IRegistrationOptions<TOptions, TCapability>` and all interfaces have been updated to implement this interface if they have both a client capability and a server registration option.

## Abstract Handlers Registration Options
The abstract base classes by the above change now have a new method that must be implemented.

`protected internal abstract TRegistrationOptions CreateRegistrationOptions(TCapability capability);`

The base handler then caches the resulting value so this method is only ever called once.

Existing implementations of `GetRegistrationOptions` can be replaced with `CreateRegistrationOptions`.

## Abstract Handlers
All the abstract handler base classes have been renamed to have a Suffix of "Base" if there was not one there before.

The fix for this is simply to make sure the `Base` suffix has been applied.

## Feature Specific Files
This is not a breaking change, the goal was to make sure that namespaces did not change, if there is a miss let me know.

Any classes related to the Request, Response, Capability and Registration Options have been moved into a `XyzFeature.cs` file.  This will help reduce the number of overall files in the repo, as well as hopefully with discoverabiliy.  More importantly this make it easier to create new features, as all the components for a given feature will be in the same file.

## TextDocumentSyncRegistrationOptions Added
The IDidXyzTextDocument handlers have been updated to have specific named options, and a master options class `TextDocumentSyncRegistrationOptions` has been added.  You provide the `TextDocumentSyncRegistrationOptions` and all the other named options can be implicitly created from it.

## Source Generation
Source Generation has been refactored to be a little more flexible and better organized (it's not prefect but better).

Some [documentation has been added](https://github.com/OmniSharp/csharp-language-server-protocol/tree/master/docs/source-generation.md) to document the attributes and their usage.  These should available to consumers to also use themselves, please let me know of any bugs you might encounter with them.


david-driscoll:
Docs link for this PR: (The above description will be rolled into the release notes).

https://github.com/OmniSharp/csharp-language-server-protocol/blob/feature/2-registration-options/docs/source-generation.md

PR #488, opened 2020-12-26
Fix invalid cast issue in omnisharp-roslyn by fixing LspHandlerDescriptor ctor

razzmatazz:
See:
 - https://github.com/OmniSharp/omnisharp-roslyn/issues/1995

The issue (apparently) is that `LspHandlerDescriptor` is constructed for `DelegatingRequest` that is registered in `LanguageServerHost.cs` (in omnisharp-roslyn) via `r.OnRequest()` here:

https://github.com/OmniSharp/omnisharp-roslyn/blob/b68f146a649bd6b0294894831e08bbd3b802110c/src/OmniSharp.LanguageServerProtocol/LanguageServerHost.cs#L310

The constructor attempts to resolve RPC return type based on what generic type is set on `IRequest<T>` that is inherited by `DelegatingRequest`. However `DelegatingRequest` implements both `IRequest` and `IRequest<TResponse>` and LINQ code selects `IRequest<MediatR.Unit>` type as the return value -- while it should be `IRequest<JToken>` for LSP requests (apparently?).

The issue was fixed by ordering implemented `IRequest<>` interfaces by generic param type and taking the first that is not `MediatR.Unit` -- only then use the `MediatR.Unit` version.

---

I am not 100% sure this is the proper/correct fix for my issue (https://github.com/OmniSharp/omnisharp-roslyn/issues/1995) but it seems to do the job.

I can try to write tests or change this you feel that is the wrong approach to fix the issue.

### DIFF
--- a/src/Protocol/Shared/LspHandlerDescriptor.cs
+++ b/src/Protocol/Shared/LspHandlerDescriptor.cs
@@ -69,8 +69,11 @@ namespace OmniSharp.Extensions.LanguageServer.Protocol.Shared
             CapabilityType = capabilityType;
 
             Response = @params?.GetInterfaces()
-                               .FirstOrDefault(x => x.IsGenericType && x.GetGenericTypeDefinition() == typeof(IRequest<>))
-                              ?.GetGenericArguments()[0] ?? typeDescriptor?.ResponseType ?? typeof(Unit);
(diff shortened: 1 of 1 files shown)


david-driscoll:
@razzmatazz This change makes sense to me, we could probably just filter `Unit` out entirely, however this should have a similar effect.

I'm not too worried about tests as the other tests have all passed.  I might look to see if I can add some more validation tests around the delegated request/responses however for the future.

Thanks!

PR #509, opened 2021-01-28
Minor updates.

david-driscoll:
* Added Container.From
* Do not serialize out the PrivateHandlerId if it's empty
* StringOrMarkupContent implict conversion should be nullable.

cc @NTaylorMullen @333fred

### DIFF
--- a/src/Protocol/Models/StringOrMarkupContent.cs
+++ b/src/Protocol/Models/StringOrMarkupContent.cs
@@ -4,7 +4,7 @@ using OmniSharp.Extensions.LanguageServer.Protocol.Serialization.Converters;
 
 namespace OmniSharp.Extensions.LanguageServer.Protocol.Models
 {
-    [JsonConverter(typeof(StringOrMarkupContentConverter))]
+    [JsonConverter(typeof( StringOrMarkupContentConverter))]
     [DebuggerDisplay("{" + nameof(DebuggerDisplay) + ",nq}")]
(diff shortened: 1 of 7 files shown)


src/Protocol/Models/StringOrMarkupContent.cs
```diff
@@ -17,9 +17,9 @@ public record StringOrMarkupContent
         public MarkupContent? MarkupContent { get; }
         public bool HasMarkupContent => String == null;
 
-        public static implicit operator StringOrMarkupContent(string value) => new StringOrMarkupContent(value);
+        public static implicit operator StringOrMarkupContent?(string? value) => value is null ? null : new StringOrMarkupContent(value);
```
333fred, line 20: Could have `NotNullIfNotNull`

333fred (approved):

NTaylorMullen (approved):
Excited to have a variant that doesn't require the serialization of the resolver handler ID, /cc @ToddGrun since we'll both probably want to adopt this for perf reasons.

src/Protocol/Models/StringOrMarkupContent.cs
333fred: @david-driscoll doesn't look like this was addressed.

333fred: @david-driscoll doesn't look like this was addressed.

PR #510, opened 2021-01-30
fix gittools actions to use the latest tagged release, the master branch as been renamed to main which caused the break

david-driscoll:
…nch as been renamed to main which caused the break

### DIFF
--- a/.github/workflows/close-milestone.yml
+++ b/.github/workflows/close-milestone.yml
@@ -13,18 +13,18 @@ jobs:
           fetch-depth: 0
 
       - name: Install GitVersion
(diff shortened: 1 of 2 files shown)


razzmatazz:
Would it be feasibile to have this release (v0.19.0-beta5) marked as proper release (it is a [pre-release](https://github.com/OmniSharp/csharp-language-server-protocol/releases/tag/v0.19.0-beta.5) now)?

So we can integrate into omnisharp and have omnisharp protocol interoperability code working via LSP?

I noticed this release is still marked as "pre-release" and https://www.nuget.org/packages/OmniSharp.Extensions.LanguageServer/ does not contain an update to -beta5.

PR #553, opened 2021-04-01
WIP tracing support

david-driscoll:
Added [tracing support](https://www.w3.org/TR/trace-context/#header-name) with help from StreamJson RPC


NTaylorMullen:
Very cool! It's nice to see under the hood and see how this stuff stitches itself together

src/Dap.Protocol/DapReceiver.cs
```diff
@@ -81,16 +83,29 @@ protected virtual IEnumerable<Renor> GetRenor(JToken @object)
...
+                        };
+                        yield break;
+                    }
+                }
 
-                    yield return new Request(sequence, RequestNames.Cancel, ro);
+                {
+                    yield return new Request(sequence, requestName, requestObject) {
+                        TraceState = request.TryGetValue("tracestate", out var ts) ? ts.Value<string>() : null,
+                        TraceParent = request.TryGetValue("traceparent", out var tp) ? tp.Value<string>() : null,
```
AArnott, line 105: Per [the spec](https://www.w3.org/TR/trace-context/#header-name), this must be a case insensitive header name match. Is that how `request.TryGetValue` is implemented?

src/Dap.Protocol/DebugAdapterConverters/DapClientNotificationConverter.cs
```diff
@@ -33,6 +33,16 @@ public override void WriteJson(JsonWriter writer, OutgoingNotification value, Js
                 writer.WritePropertyName("body");
                 serializer.Serialize(writer, value.Params);
             }
+            if (value.TraceParent != null)
+            {
+                writer.WritePropertyName("traceparent");
+                writer.WriteValue(value.TraceParent);
+            }
+            if (value.TraceState != null)
```
AArnott, line 41: [The spec](https://www.w3.org/TR/trace-context/#tracestate-header-field-values) discourages sending the `tracestate` header with an empty value. Should this condition be changed to `!string.IsNullOrEmpty(value.TraceState)`?

AArnott: Also, this block should be *inside* the `if (value.TraceParent != null)` block to satisfy [the spec](https://www.w3.org/TR/trace-context/#no-traceparent-received) that says `tracestate` without `traceparent` is invalid.

AArnott (approved):
Looks good. I'm glad the vs-streamjsonrpc code helped expedite this.

src/Dap.Protocol/DapReceiver.cs
david-driscoll: Thanks!

src/Dap.Protocol/DebugAdapterConverters/DapClientNotificationConverter.cs
david-driscoll: Thanks!

AArnott:
If I had $1 for every PR that completed with "WIP" in the title...

Issue #562, opened 2021-04-13
Unable to start Debug Server: Constructor on type [...] not found


rjmholt:
Trying to get PowerShell Editor Services running with O# 0.19, we've been unable to start the debug server.

In our code, the failure seems to occur [here](https://github.com/PowerShell/PowerShellEditorServices/blob/7c035bcda4cacfbb8060548468daf615934fed61/src/PowerShellEditorServices/Server/PsesDebugServer.cs#L76).

It seems we're unable to process the `initialize` message from the debugger because the `Activator` can't create the `PathFormat` type:

https://github.com/OmniSharp/csharp-language-server-protocol/blob/1b6788df2600083c28811913a221ccac7b1d72c9/src/Dap.Protocol/Models/PathFormat.cs#L5-L10

```
Constructor on type 'System.Nullable`1[[OmniSharp.Extensions.DebugAdapter.Protocol.Models.PathFormat, OmniSharp.Extensions.DebugAdapter, Version=0.19.0.0, Culture=neutral, PublicKeyToken=6d868dff454e6022]]' not found.
   at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture) in /_/src/coreclr/System.Private.CoreLib/src/System/RuntimeType.CoreCLR.cs:line 3541
   at System.Activator.CreateInstance(Type type, Object[] args) in /_/src/libraries/System.Private.CoreLib/src/System/Activator.cs:line 31
   at OmniSharp.Extensions.JsonRpc.Serialization.Converters.EnumLikeStringConverter.ReadJson(JsonReader reader, Type objectType, IEnumLikeString existingValue, Boolean hasExistingValue, JsonSerializer serializer) in /_/src/JsonRpc/Serialization/Converters/EnumLikeStringConverter.cs:line 17
   at Newtonsoft.Json.JsonConverter1.ReadJson(JsonReader reader, Type objectType, Object existingValue, JsonSerializer serializer) in /_/Src/Newtonsoft.Json/JsonConverter.cs:line 126
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.DeserializeConvertable(JsonConverter converter, JsonReader reader, Type objectType, Object existingValue) in /_/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs:line 2175
[...]

PR #563, opened 2021-04-15
Fix nullable struct enum deserialization

rjmholt:
Fixes https://github.com/OmniSharp/csharp-language-server-protocol/issues/562.

Not sure if this is the ideal fix, just a possible contribution 🙂


rjmholt:
Also not sure where a good place to add a test would be

PR #564, opened 2021-04-15
Fixed a bug where nullable enum like classes could not be deserialized correctly

david-driscoll:


rjmholt (approved):

rjmholt:
Thanks @david-driscoll!

rjmholt:
Fixes https://github.com/OmniSharp/csharp-language-server-protocol/issues/562

src/JsonRpc/Serialization/Converters/EnumLikeStringConverter.cs
```diff
@@ -13,9 +13,10 @@ internal class EnumLikeStringConverter : JsonConverter<IEnumLikeString>
             bool hasExistingValue,
             JsonSerializer serializer
         ) =>
-            reader.TokenType switch {
-                JsonToken.String => (IEnumLikeString) Activator.CreateInstance(objectType, (string) reader.Value),
-                _                => (IEnumLikeString) Activator.CreateInstance(objectType, null)
+            ( reader.TokenType, Nullable.GetUnderlyingType(objectType) ) switch {
+                (JsonToken.String, null)         => (IEnumLikeString) Activator.CreateInstance(objectType, (string) reader.Value),
+                (JsonToken.String, { } realType) => (IEnumLikeString) Activator.CreateInstance(realType, (string) reader.Value),
+                _                                => (IEnumLikeString) Activator.CreateInstance(objectType, null)
```
rjmholt, line 19: If we get something like `PathFormat?` (implying the value can be `null`), we'll see a `null` from `reader.Value` I assume.

I would have expected the serialiser to return `null` in that circumstance rather than `new PathFormat(null)`.

But also, if we do get a `PathFormat?`, we can't promise a non-nullable `IEnumLikeString` but rather I would have guessed we need to return `IEnumLikeString?`.

So in the `realType` branch, I would have expected something like:

```
reader.Value is null ? null : (IEnumLikeString) Activator.CreateInstance(realType, (string) reader.Value)
```

Does that make sense? Or is there something else going on here that makes that not a concern?

david-driscoll: that sounds correct, I'll update the pattern

Issue #566, opened 2021-04-17
Do the threads returned by a ThreadsHandler need to be real threads?


rjmholt:
Today, a `ThreadsHandler` returns a `Container<System.Threading.Thread>` and debugger events send through the ID of the affected thread.

In a world where most of our backend is async we don't really have a specific thread for anything we're debugging (in PowerShell we technically do, but it's kind of hard to get to).

So if we want to send back a thread through the ThreadsHandler, the easiest thing to do is to create a dummy thread object to represent the debugged thread -- but if that's the case, we might be better off with an O#-defined object to represent a thread.

Today we send back an empty container and always set `AllThreadsStopped` and `AllThreadsContinued`, but that might not be a good way to go...?

rjmholt:
In particular I notice there's a type called `Omnisharp.Extensions.DebugAdapter.Protocol.Models.Thread` that seems appropriate for this

yevgeni-zolotko:
I also noticed the same issue. I think this is a bug in a code generator which confuses `Omnisharp.Extensions.DebugAdapter.Protocol.Models.Thread` with `System.Threading.Thread`.
This leads to a JSON serialization exception, so using the `threads` request crashes the application.

PR #567, opened 2021-04-19
Fix thread debugging

rjmholt:
Fixes https://github.com/OmniSharp/csharp-language-server-protocol/issues/566


src/Dap.Protocol/Feature/Requests/ThreadsFeature.cs
```diff
@@ -2,7 +2,6 @@
 using OmniSharp.Extensions.DebugAdapter.Protocol.Models;
 using OmniSharp.Extensions.JsonRpc;
 using OmniSharp.Extensions.JsonRpc.Generation;
-using Thread = System.Threading.Thread;
```
andschwa, line 5: OMG this is all it was?! So this type came in and overrode the Thread in `Protocol.Models`...I was a bit confused by this when I brought in the update but didn't know how much of a worry it was: https://github.com/PowerShell/PowerShellEditorServices/pull/1424/files#r616249975

david-driscoll:
Thanks @rjmholt

PR #633, opened 2021-08-10
Treat missing `.ConfigureAwait(false)` as errors (and fix)

andschwa:
I enabled the new .NET Analyzer for the project and then used an Editor Config rule to treat CA2007 as an error. This diagnostic is for any awaited task that does not have an explicit trailing `.ConfigureAwait(false)` call, an annoying but very necessary configuration for asynchronous .NET library code (such as OmniSharp).

The bugs caused by these missing calls are strange. In the case of PowerShell Editor Services, an LSP server using OmniSharp and powering the PowerShell Extension for VS Code, it showed up as a hang when the server executed user code that used objects from `System.Windows.Forms`.

This is because that .NET library sets its own synchronization context, which is exactly where `ConfigureAwait(continueOnCapturedContext: false)` comes into play. The default value is `true` which roughly means that the awaited tasks will only run on their own context. So when the context is changed (because of `System.Windows.Forms` or other code introducing a new synchronization context) the task will never be continued, resulting in this hang. By configuring this to `false` we allow the tasks to continue regardless of the new context.

See this .NET blog post for more details: https://devblogs.microsoft.com/dotnet/configureawait-faq/

Note that elsewhere in the codebase we've been very careful to set it to false, but we've not been perfect. Treating this diagnostic as an error will allow us to be as perfect as possible about it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -128,7 +128,11 @@ resharper_web_config_module_not_resolved_highlighting=warning
 resharper_web_config_type_not_resolved_highlighting=warning
 resharper_web_config_wrong_module_highlighting=warning
 
+# .NET Analzyer settings
+# VSTHRD200: Use "Async" suffix for awaitable methods
 dotnet_diagnostic.VSTHRD200.severity = none
+# CA2007: Do not directly await a Task
+dotnet_diagnostic.CA2007.severity = error
 
 [*.{cs,cshtml}]
 charset=utf-8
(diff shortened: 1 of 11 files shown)


andschwa:
I've had users confirm that this change solved https://github.com/PowerShell/vscode-powershell/issues/3394, https://github.com/PowerShell/vscode-powershell/issues/3410, and https://github.com/PowerShell/vscode-powershell/issues/3415. Note that I first applied this change to the PowerShell Editor Services code base in https://github.com/PowerShell/PowerShellEditorServices/pull/1533, which alone did not solve the issue.

The same bug was found and fixed in the past with https://github.com/PowerShell/PowerShellEditorServices/pull/1149 which gave me the idea.

andschwa:
Ah, I need to fix the tests too, I guess those didn't build from the build script.

rjmholt (approved):

andschwa:
Hey, it's not mysterious! It's a threading deadlock 😆

PR #635, opened 2021-08-12
Fix potential unobserved exception in InputHandler.cs

ycholette:
Observable.Do() does not catch exceptions from FromAsync(), resulting in a potential unobserved exception.

### DIFF
--- a/src/JsonRpc/InputHandler.cs
+++ b/src/JsonRpc/InputHandler.cs
@@ -110,9 +110,16 @@ namespace OmniSharp.Extensions.JsonRpc
         public void Start()
         {
             _disposable.Add(
-                Observable.FromAsync(() => ProcessInputStream(_stopProcessing.Token))
-                          .Do(_ => { }, e => _logger.LogCritical(e, "unhandled exception"))
-                          .Subscribe(_inputActive)
+                Observable.FromAsync(async () => {
+                    try
+                    {
+                        await ProcessInputStream(_stopProcessing.Token).ConfigureAwait(false);
+                    }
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OmniSharp/csharp-language-server-protocol?pullRequest=635) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OmniSharp/csharp-language-server-protocol?pullRequest=635) before we can accept your contribution.<br/><br/>:x: ycholette [sign now](https://cla.dotnetfoundation.org/OmniSharp/csharp-language-server-protocol?pullRequest=635)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OmniSharp/csharp-language-server-protocol?pullRequest=635) it.</sub>

david-driscoll:
Is there an example case where you're seeing this as an issue?

Is there any docs or source you can point to?  I would have assumed any exception would just go to the error flow.

ycholette:
I think this is related to this: https://github.com/dotnet/reactive/issues/1256
Basically, the cancelation token was cancelled, `_pipeReader.AdvanceTo(buffer.Start, buffer.End)` threw an exception (before the check on the token), and the exception was not observed by anyone. My break point in the Do() never hit.

david-driscoll:
Alright, that makes a little more sense! Thanks!

Issue #609, opened 2021-07-01
Server unhandled exception when trying to parse `traceparent` property in `Initialize` message. Client (VS2019) becomes unresponsive.


sstublic:
I'm not sure if I'm missing something obvious, but my scenario is as follows:

Running `OmniSharp LanguageServer 0.19.2` and `VS 2019 version 16.10.3`. When `Initialize` message is sent, the server reports an error:
```
OmniSharp.Extensions.JsonRpc.InputProcessingException: There was an error processing input the contents of the buffer were ''
 ---> System.InvalidCastException: Cannot cast Newtonsoft.Json.Linq.JProperty to Newtonsoft.Json.Linq.JToken.
   at Newtonsoft.Json.Linq.Extensions.Convert[T,U](T token)
   at Newtonsoft.Json.Linq.Extensions.Value[T,U](IEnumerable`1 value)
   at Newtonsoft.Json.Linq.Extensions.Value[U](IEnumerable`1 value)
   at OmniSharp.Extensions.JsonRpc.Receiver.GetRenor(JToken object)
   at OmniSharp.Extensions.JsonRpc.Receiver.GetRequests(JToken container)
   at OmniSharp.Extensions.LanguageServer.Server.LspServerReceiver.GetRequests(JToken container)
   at OmniSharp.Extensions.JsonRpc.InputHandler.HandleRequest(ReadOnlySequence`1& request)
   at OmniSharp.Extensions.JsonRpc.InputHandler.ProcessInputStream(CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
```

After this error VS2019 becomes unresponsive and server doesn't log additional events.

The culprit seems to be the line https://github.com/OmniSharp/csharp-language-server-protocol/blob/f5c9691db0540c6c5870d3eba4ad4bd538852f25/src/JsonRpc/Receiver.cs#L116

VS2019 will provide `traceparent` value but the above line will always fail with reported exception if there is any value in the property.

It seems to me that line should read: 

`TraceParent = properties["traceparent"].FirstOrDefault()?.Value.Value<string>()`

instead of 

`TraceParent = properties["traceparent"].FirstOrDefault()?.Value<string>()`

However, this would mean that with addition of `traceparent`/`tracestate` property parsing NO clients which actually send those values would work at all with `OmniSharp 0.19.2`. That seems unlikely.

Any ideas what I'm missing?

Thanks!

NTaylorMullen:
[...]

PR #638, opened 2021-08-17
Fix trace state parsing.

NTaylorMullen:
- Without this the `.Value<string>` was exploding attempting to convert to a string.

### DIFF
--- a/src/JsonRpc/Receiver.cs
+++ b/src/JsonRpc/Receiver.cs
@@ -100,20 +100,25 @@ namespace OmniSharp.Extensions.JsonRpc
 
             var properties = request.Properties().ToLookup(z => z.Name, StringComparer.OrdinalIgnoreCase);
 
+            var traceStateProperty = properties["tracestate"].FirstOrDefault();
+            var traceState = traceStateProperty?.Value.ToString();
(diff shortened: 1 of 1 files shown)


andschwa:
@NTaylorMullen just out of curiosity, can I see how this was failing for you? In case something similar crops up in PowerShell Editor Services 😆

NTaylorMullen:
> @NTaylorMullen just out of curiosity, can I see how this was failing for you? In case something similar crops up in PowerShell Editor Services 😆

I have it all patched locally so not easy to reproduce but it was a Cast exception. The gist is that when you got a jsonrpc payload which had:
```
{
    "traceparent": "alskdfj-asldkjf-asdj-asdflkj-sadfjasd-fj"
    ....
}
```

If you have the powershell language server running in VS you'll see the exception too (they provide tracing); although I don't believe VSCode's jsonrpc bits do.

andschwa:
Oh interesting, not sure I was using `traceparent` yet but I remember seeing it get enabled.

andschwa:
Ah yes this PR: #553

andschwa:
Taylor, do you think your PR will fix #609 too?

NTaylorMullen:
> Taylor, do you think your PR will fix #609 too?

Yup, absolutely will fix that!

PR #641, opened 2021-08-23
Refactor request invocation to be extensible.

NTaylorMullen:
- The initial motivation for this change spawned from an investigation in the Razor language server where transitioning from Parallel -> Serial tasks would wait on Parallel work to finish. In scenarios when that parallel work took a long time this could result in significant editor delays in completion where you'd have Parallel (long) -> Serial -> Parallel (completion, short). I played around with changing the System.Reactive bits to have an option to not "wait" for the parallel stacks but System.Reatcive as a library wasn't truly built to handle that type of "change your mind after-the-fact" flow.
- Prior to this change the routing & scheduling aspects of the JsonRpc stack are bound to our ProcessScheduler & InputHandler.RouteRequest & InputHandler.RouteNotification endspoints. This change allows that entire stack to be extensible so consumers can plug & play.
- Added a `RequestInvoker` type which represents the core logic of how the framework invokes a handler for a request. This encapsulates the control flow for invoking, scheduling and handling fallout from invoking a handler.
- Added a `RequestInvokerOptions` type to represent what sort of settings should be applied for the request invoker paradigm.
- Expanded `InputHandler` & `Connection` to have two new constructors that take in a request invoker and obsoleted the old ones. Updated tests to account for this.
- Registered the default request invoker type (the one that uses System.Reactive) if a request invoker was not already registered.

Make existing request response types fully public
- For consumers who are creating their own `RequestInvoker` they need to manually construct many of our response types. Therefore, the constructors need to also be puclic.

### DIFF
--- a/src/JsonRpc/DefaultRequestInvoker.cs
+++ b/src/JsonRpc/DefaultRequestInvoker.cs
@@ -1,0 +1,181 @@
+using System;
+using System.Reactive;
+using System.Reactive.Concurrency;
+using System.Reactive.Disposables;
+using System.Reactive.Linq;
+using Microsoft.Extensions.Logging;
+using OmniSharp.Extensions.JsonRpc.Server;
+using OmniSharp.Extensions.JsonRpc.Server.Messages;
+using Notification = OmniSharp.Extensions.JsonRpc.Server.Notification;
+
+namespace OmniSharp.Extensions.JsonRpc
+{
+    public class DefaultRequestInvoker : RequestInvoker
+    {
+        private readonly IRequestRouter<IHandlerDescriptor?> _requestRouter;
+        private readonly IOutputHandler _outputHandler;
+        private readonly ProcessScheduler _processScheduler;
+        private readonly IRequestProcessIdentifier _requestProcessIdentifier;
+        private readonly RequestInvokerOptions _options;
+        private readonly ILogger<DefaultRequestInvoker> _logger;
+
+        public DefaultRequestInvoker(
+            IRequestRouter<IHandlerDescriptor?> requestRouter,
+            IOutputHandler outputHandler,
+            IRequestProcessIdentifier requestProcessIdentifier,
+            RequestInvokerOptions options,
+            ILoggerFactory loggerFactory,
+            IScheduler scheduler)
+        {
+            _requestRouter = requestRouter;
+            _outputHandler = outputHandler;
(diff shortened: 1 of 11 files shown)


src/JsonRpc/DefaultRequestInvoker.cs
```diff
@@ -0,0 +1,181 @@
...
+            IRequestDescriptor<IHandlerDescriptor?> descriptor,
+            Request request,
+            RequestInvocationHandle handle)
+        {
+            var cts = handle.CancellationTokenSource;
+            return (contentModifiedToken, scheduler) =>
+                Observable.Create<ErrorResponse>(
+                               observer => {
+                                   // ITS A RACE!
+                                   var sub = Observable.Amb(
```
NTaylorMullen, line 81: All this is verbatim from the original RouteRequest method in InputHandler.

src/JsonRpc/JsonRpcServerServiceCollectionExtensions.cs
```diff
@@ -47,13 +47,22 @@ public static class JsonRpcServerServiceCollectionExtensions
...
+            container.Register<RequestInvokerOptions>(
+                made: new Made.TypedMade<RequestInvokerOptions>().Parameters
+                                                      .Type<TimeSpan>(serviceKey: nameof(options.MaximumRequestTimeout))
+                                                      .Type<bool>(serviceKey: nameof(options.SupportsContentModified))
+                                                      .Name("concurrency", serviceKey: nameof(options.Concurrency)),
+                reuse: Reuse.Singleton);
+
+            if (!container.IsRegistered<RequestInvoker>())
+            {
+                container.Register<RequestInvoker, DefaultRequestInvoker>(Reuse.Singleton);
```
david-driscoll, line 60: This is missing `.Type<IScheduler>(serviceKey: nameof(options.InputScheduler))`

david-driscoll (changes requested):
LGTM -- Just the one issue with the input scheduler, and then it should be GTG.

src/JsonRpc/JsonRpcServerServiceCollectionExtensions.cs
NTaylorMullen: ooo, out of curiosity why not have that be registered by default?

NTaylorMullen:
🆙 📅

andschwa:
Very cool work here!

Issue #676, opened 2021-10-31
MediatR v9.0.0 includes a breaking change.


chrissimon-au:
This is probably an issue that is not going to affect many people, but I believe it is a legit dependency issue that will have to be resolved at some point:

`Directory.Build.targets` includes `MediatR` at version `8.1.0`.

When adding as a dependency via dotnet or nuget, (e.g. `dotnet add package OmniSharp.Extensions.LanguageServer`), which is what I'd guess most folk are doing, it follows the [Nuget Versioning](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges) rules and pulls in the lowest version >= the specified version - which resolves to `8.1.0` and works fine.

When adding via `paket add OmniSharp.Extensions.LanguageServer` it follows the [paket rules](https://fsprojects.github.io/Paket/controlling-nuget-resolution.html) which identifies the _latest matching version_ which is `9.0.0` and unfortunately this version has a breaking change.  The issue is only detectable at runtime due to the use of reflection I believe.  The problem is an attempt to invoke a method that no longer exists on the `IMediator` interface, due to being moved to a separate interface that `IMediator` implements. 

Compare:

v8.1.0: https://github.com/jbogard/MediatR/blob/v8.1.0/src/MediatR/IMediator.cs

vs

v9.0.0: https://github.com/jbogard/MediatR/blob/v9.0.0/src/MediatR/IMediator.cs

The exception that's thrown is:

```
OmniSharp.Extensions.JsonRpc.DefaultRequestInvoker: Failed to handle request initialize 1 -
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.MissingMethodException: Method not found: 'System.Threading.Tasks.Task`1<System.__Canon> MediatR.IMediator.Send(MediatR.IRequest`1<System.__Canon>, System.Threading.CancellationToken)'.
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
[...]

PR #678, opened 2021-10-31
Update MediatR to 9.0.0

chrissimon-au:
A possible fix for #676

Sets the MediatR version using the version range syntax to `[9.0.0,10.0.0)` (as a precaution for future breaking changes in MediatR v10.

Have done a local build and all tests pass without any changes.  It seems that just building against the explicit 9.0.0 dependency is sufficient to ensure the correct mediator interface method is invoked.


JoeRobich:
Took update in https://github.com/OmniSharp/csharp-language-server-protocol/pull/1084

PR #774, opened 2022-03-15
Update lsp.md

LazyTaz:
"oin" -> "on"

Thanks for the amazing work !


david-driscoll:
not sure why the build is not triggering as expected, going to mess with this.

PR #892, opened 2022-10-05
Fix bug caused by `System.Threading.Tasks` v4.7.1

andschwa:
Unfortunately `System.Threading.Tasks` v4.7.1 is broken. Its package manifest states that it depends on `System.Threading.Tasks.Extensions` >= v4.5.4 (assembly version v4.2.0.1); however, for some reason it actually requests v4.2.0.0. While neither OmniSharp nor PowerShell Editor Services, the downstream library where this bug appeared, ship that version of the library, it can sometimes be found in the Global Assembly Cache on Windows. When present (due to an installation by some other program), it gets loaded, and then other dependencies (correctly) request v4.2.0.1 which also gets loaded. When both are loaded, a `System.MissingMethodException` is thrown! Specifically the method `ChannelReader.ReadAsync` (from `System.Threading.Tasks`) on the type `Tasks.ValueTask` (provided by `System.Threading.Tasks.Extensions`) can no longer be resolved. I do not know enough about the intricacies of method/type resolution in the full .NET Framework as to explain precisely why the presence of v4.2.0.0 causes this resolution to fail, but it certainly does. Upgrading `System.Threading.Tasks` to v6.0.0 is confirmed to solve the problem, since the newer package correctly requests `System.Threading.Tasks.Extensions` v4.2.0.1.

I tested this and it sure seems to fix https://github.com/PowerShell/vscode-powershell/issues/4175.

### DIFF
--- a/.build/.build.csproj
+++ b/.build/.build.csproj
@@ -14,7 +14,6 @@
         <PackageReference Include="JetBrains.ReSharper.CommandLineTools" ExcludeAssets="All" />
         <PackageReference Include="GitVersion.Tool" ExcludeAssets="All" />
         <PackageReference Include="ReportGenerator" ExcludeAssets="All" />
-        <PackageReference Include="JetBrains.ReSharper.CommandLineTools" ExcludeAssets="All" />
         <PackageReference Include="Rocket.Surgery.Nuke" />
     </ItemGroup>
 
--- a/sample/SampleServer/SampleServer.csproj
+++ b/sample/SampleServer/SampleServer.csproj
@@ -10,7 +10,7 @@
 
     <ItemGroup>
         <ProjectReference Include="../../src/Server/Server.csproj" />
-        <PackageReference Include="Microsoft.Extensions.Logging" VersionOverride="3.0.0" />
+        <PackageReference Include="Microsoft.Extensions.Logging" VersionOverride="6.0.0" />
         <PackageReference Include="Serilog.Extensions.Logging" VersionOverride="3.1.0" />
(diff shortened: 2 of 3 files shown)


.build/.build.csproj
```diff
@@ -14,7 +14,6 @@
         <PackageReference Include="JetBrains.ReSharper.CommandLineTools" ExcludeAssets="All" />
         <PackageReference Include="GitVersion.Tool" ExcludeAssets="All" />
         <PackageReference Include="ReportGenerator" ExcludeAssets="All" />
-        <PackageReference Include="JetBrains.ReSharper.CommandLineTools" ExcludeAssets="All" />
```
andschwa, line 17: I removed this because it was a duplicate line and the build tools complained about it. (See line 14.)

sample/SampleServer/SampleServer.csproj
```diff
@@ -10,7 +10,7 @@
 
     <ItemGroup>
         <ProjectReference Include="../../src/Server/Server.csproj" />
-        <PackageReference Include="Microsoft.Extensions.Logging" VersionOverride="3.0.0" />
```
andschwa, line 13: The sample project was failing to restore because its dependencies updated to 6.0.0, but it was not.

andschwa:
@david-driscoll I split the changes into three commits in case you decide to release v0.19.6 with just a backported patch, since I'm not sure how ready it is for a full release.

andschwa:
I've also confirmed with @prpercival that the dev build I produced with the updated O# package fixed the issue on their machine too 😄

andschwa:
@david-driscoll the bot added the "mysterious" label and that is _so freaking apt_!

PR #955, opened 2023-03-17
Don't log critical message when client disconnect

ycholette:
We are showing warnings/errors in our UI and this would pop up every time a client disconnects (usually by the user). This error should not result in a critical message in the log IMO. What do you think?  
  
Not 100% about the wording of the message to show in this case though.


JoeRobich (approved):
Thanks @ycholette! I think this change makes sense.

PR #1049, opened 2023-10-02
Add npm to CI build

manandre:
Should avoid introducing breaking changes in the `vscode-testextension` project and int its `npm` dependencies in particular.


JoeRobich:
So, for C# extension integration tests we needed to setup a virtual display for our linux leg and pass the id in the `DISPLAY` environment variable. See https://github.com/dotnet/vscode-csharp/blob/ed352248c09fa57cfb2c2c0b04874fd122a26538/.github/workflows/ci.yml#L18

JoeRobich (approved):
Thanks @manandre!

Issue #1037, opened 2023-09-26
Suspected Memory Leak in Language Client


antpensi:
Hello

I am investigating a suspected memory leak in the Omnisharp Language Client. When configuring the client to handle a periodic notification from an external Language Server, I consistently observe memory usage growing unbounded.

The memory profiler within VS indicates this is coming from many Newtonsoft.Json.Linq.* objects related to the OmniSharp.Extensions.JsonRpc.Server.Notification class.

These objects appear to grow even when constructing a "no-op" notification handler, which leads me to believe it may be internal to this library.

Here is how I am constructing my Language Client - 

```csharp
Process process = new Process(...configured for myLangServer.exe...);

LanguageClientOptions options = new LanguageClientOptions();
options.WithInput(process.StandardOutput.BaseStream)
       .WithOutput(process.StandardInput.BaseStream)
       .OnNotification("foo/bar", (MyNotification notification) =>
       {
	    Debug.Print($"Notification received.");
       });

var languageClient = OmniSharp.Extensions.LanguageServer.Client.LanguageClient.Create(options);
await languageClient.Initialize(CancellationToken.None);
```

I am using version 0.19.7 of the Omnisharp.Extensions.LanguageClient NuGet package. Attaching an image detailing an example of one of these Newtonsoft objects.

![devenv_PaNuYVndH7](https://github.com/OmniSharp/csharp-language-server-protocol/assets/7108065/b84fb22f-04a6-4b1a-b84a-10ad50ac40d6)

antpensi:
![image](https://github.com/OmniSharp/csharp-language-server-protocol/assets/7108065/fbea05e9-b2b9-460c-867c-de6f63f40cbd)

PR #1053, opened 2023-10-04
Fix ProcessScheduler memleak

antpensi:
Reducing the int.MaxValue here to an arbitrary low int value seems to alleviate the suspected memory leak described in https://github.com/OmniSharp/csharp-language-server-protocol/issues/1037 

Hoping this PR can serve as the start of a conversation about what might be a more proper solution.

### DIFF
--- a/src/JsonRpc/ProcessScheduler.cs
+++ b/src/JsonRpc/ProcessScheduler.cs
@@ -34,7 +34,7 @@ namespace OmniSharp.Extensions.JsonRpc
 
                     var observableQueue =
                         new BehaviorSubject<(RequestProcessType type, ReplaySubject<IObservable<Unit>> observer, Subject<Unit>? contentModifiedSource)>(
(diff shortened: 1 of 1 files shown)


antpensi:
@microsoft-github-policy-service agree company="VMware"

antpensi:
Hi @JoeRobich - please excuse the direct ping. Do you happen to know the right person to get eyes on this issue? I do not necessarily think the changes in this PR are the correct solution, but it does seem to alleviate the memory leak described in #1037. Hope to use this as a conversation point.

Thank you

JoeRobich:
Hi @antpensi - I am likely the right person, thanks for the ping. 

I saw this come up earlier and still need to investigate how the ReplaySubject is being used. Am I right that it captures requests while a serial request is being processed, so that they can be replayed when the blocking request is complete? Why is 6 seconds long enough? Does a replay subject make the most sense here or would a queue work just as well?

PR #1077, opened 2023-10-20
Fix NotebookSelector type

### LINKED ISSUE #1066: Notebook Document support is not implemented correctly
Hello,

I was trying to connect from the LanguageClient to the **pylsp** server and when I did this with the current version of pylsp that supports LSP v. 3.17, the `Initialize` call ends in following Exception related to JSON Deserialization:

`Cannot deserialize the current JSON array (e.g. [1,2,3]) into type 'OmniSharp.Extensions.LanguageServer.Protocol.Models.NotebookSelector' because the type requires a JSON object (e.g. {"name":"value"}) to deserialize correctly`

 I did some digging and the culprit seems to be this part of the pylsp response that specifies server capabilities:

`'notebookDocumentSync': {'notebookSelector': [{'cells': [{'language': 'python'}]}]}`

As you can see, the top-level object of notebookSelector is array, which is (if I understand the LSP documentation correctly) a valid syntax, but the LanguageClient does not seem to support it.

manandre:
Fixes #1066

The `NotebookSelector` type is an array type.
See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notebookDocumentSyncOptions

### DIFF
--- a/src/Protocol/Features/Document/NotebookDocumentSyncFeature.cs
+++ b/src/Protocol/Features/Document/NotebookDocumentSyncFeature.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


JoeRobich (approved):
Thanks @manandre!

PR #1079, opened 2023-10-22
Fix RenameFiles workspace operations

### LINKED ISSUE #1078: FileRename record implementation is incorrect
In [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#fileRename):

```typescript
export interface FileRename {
    oldUri: string;
    newUri: string;
}
```

In [protocol implementation](https://github.com/OmniSharp/csharp-language-server-protocol/blob/master/src/Protocol/Features/FileOperationsFeature.cs#L111):

```csharp
public abstract record FileOperationItem
{
    public Uri Uri { get; init; } = null!;
}

public partial record FileRename : FileOperationItem;
```

When using `DidRenameFileParams`/`WillRenameFileParams` (handling `workspace/didRenameFiles`/`workspace/willRenameFiles`), items in `request.Files` only include one URI. I haven't tested the requests but I suppose the param objects cannot be deserialized as expected.
(Packages like `vscode-languagesclient` implement this correctly.)

---

By the way, shouldn't these classes be named as _DidRenameFile**s**Params_/_WillRenameFile**s**Params_ (plural form; the same also applies to corresponding handler classes and interfaces)?

manandre:
Fixes #1078

### DIFF
--- a/test/Lsp.Integration.Tests/FileOperationsTests.cs
+++ b/test/Lsp.Integration.Tests/FileOperationsTests.cs
@@ -17,24 +17,24 @@ using Xunit.Abstractions;
 
 namespace Lsp.Integration.Tests
 {
-    public class FileOperationTests : LanguageProtocolTestBase
+    public class FileOperationsTests : LanguageProtocolTestBase
     {
-        private readonly Action<DidCreateFileParams> _didCreateFileHandler = Substitute.For<Action<DidCreateFileParams>>();
+        private readonly Action<DidCreateFilesParams> _didCreateFileHandler = Substitute.For<Action<DidCreateFilesParams>>();
 
-        private readonly Func<WillCreateFileParams, Task<WorkspaceEdit?>> _willCreateFileHandler =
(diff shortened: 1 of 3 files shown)


test/Lsp.Integration.Tests/FileOperationsTests.cs
```diff
@@ -43,14 +43,14 @@ public async Task Should_Handle_FileCreate()
         {
             var (client, server) = await Initialize(Configure, Configure);
 
-            await client.RequestWillCreateFile(
-                new WillCreateFileParams
+            await client.RequestWillCreateFiles(
```
manandre, line 46: The plural form seems to impact the `ILanguageClient` extension methods. Is it considered a breaking change?

JoeRobich: I think we are fine to take the changes, but I would like us to start a breaking changes document to catalogue the breaking changes in each version.

PR #1118, opened 2023-11-19
Add ICapability to ShowMessageRequestClientCapabilities. Fixes #1117

### LINKED ISSUE #1117: `ShowMessageRequestClientCapabilities` should implement `ICapability`
When attempting to use the `LanguageServerTestBase` to construct a LanguageClient for e2e testing of my language server, I'd like to test different scenarios for the client supporting message requests and not.

Unfortunately, [ShowMessageRequestClientCapabilities](https://github.com/OmniSharp/csharp-language-server-protocol/blob/ab6e519ce22104a2f7f9c8573b0d14671b020608/src/Protocol/Features/Window/ShowMessageRequestFeature.cs#L78) doesn't implement `ICapability`, so when trying to add it to the client options with `WithCapability` it fails due to the type requirement.

This seems like a simple fix, so I'll be linking a PR shortly.

chrissimon-au:
Fixes #1117 by adding the interface `ICapability` which should allow it to be used by the `LanguageClientOptions.WithCapability` method.


JoeRobich (approved):
@chrissimon-au Thanks for the fix!

PR #1171, opened 2024-01-13
Fix github actions badge

sh-cho:
- https://github.com/badges/shields/issues/8671


sh-cho:
@microsoft-github-policy-service agree

JoeRobich (approved):
Thanks @sh-cho!

PR #1260, opened 2024-06-16
Debugging for RSG dependencies update

manandre:
Forcing verbose logging for Restore step in order to get some extra information about the NRE.
See #1120


JoeRobich:
@manandre Thanks for troubleshooting this. This PR passed because it was using an older version of GitVersion.Tool. The problem with using the newer GitVersion.Tool was with the NuGet caching in the CI. We were always pulling in an outdated cache and never replacing it. Going to close this out now.

PR #1439, opened 2025-01-29
Upgrade MediatR version to 12.4.1

sheffer21:
- Change all MediatR IRequest usages to IRequest{Unit}. MediatR doesn't use the Unit return type by default anymore. To keep similar signatures I updated all usages to IRequest{Unit}


JoeRobich (approved):
Thanks @sheffer21!

PR #1443, opened 2025-02-01
Fix serilog update

manandre:
Fixes #1410

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -28,13 +28,13 @@
   </ItemGroup>
   <ItemGroup>
     <!-- backwards compat is maintained in Directory.Packages.supports.props -->
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="8.0.1" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="8.0.1" />
-    <PackageVersion Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.2" />
-    <PackageVersion Include="Microsoft.Extensions.Options" Version="8.0.2" />
-    <PackageVersion Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="8.0.0" />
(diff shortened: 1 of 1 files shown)


Directory.Packages.props
```diff
@@ -28,13 +28,13 @@
...
   <ItemGroup>
     <!-- backwards compat is maintained in Directory.Packages.supports.props -->
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="8.0.1" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="8.0.1" />
-    <PackageVersion Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.2" />
-    <PackageVersion Include="Microsoft.Extensions.Options" Version="8.0.2" />
-    <PackageVersion Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="8.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="9.0.1" />
```
manandre, line 31: The build logs contain new scary warnings:
```
warning : Microsoft.Extensions.Logging 9.0.1 doesn't support net6.0 and has not been tested with it. Consider upgrading your TargetFramework to net8.0 or later. You may also set <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings> in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk.
```
Should we remove the support of the net6.0 TFM as it is officially EOL (see https://versionsof.net/core/6.0/)?

JoeRobich: The Microsoft.Extensions.Logging package does include .netstandard builds which should be fine. I imagine warnings like this are useful to push people to upgrade.

> Should we remove the support of the net6.0 TFM as it is officially EOL (see https://versionsof.net/core/6.0/)?

I think we should. I need to make progress on https://github.com/OmniSharp/omnisharp-roslyn/issues/2647 which would free O# to move to .net8.0 as well.

